PR #15, opened 2022-07-22
Misc minor updates

RockyMandayam2:
Misc minor updates, part of me poking around while onboarding.

### DIFF
--- a/src/main/java/com/facebook/openwifirrm/DeviceConfig.java
+++ b/src/main/java/com/facebook/openwifirrm/DeviceConfig.java
@@ -79,15 +79,6 @@ public class DeviceConfig {
 		config.enableRRM = true;
 		config.enableConfig = true;
 		config.enableWifiScan = true;
-		config.boundary = null;
-		config.location = null;
-		config.allowedChannels = null;
-		config.allowedChannelWidths = null;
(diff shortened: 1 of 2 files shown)


src/main/java/com/facebook/openwifirrm/DeviceConfig.java
```diff
@@ -79,15 +79,6 @@ public static DeviceConfig createDefault() {
...
 		config.enableWifiScan = true;
-		config.boundary = null;
-		config.location = null;
-		config.allowedChannels = null;
-		config.allowedChannelWidths = null;
-		config.autoChannels = null;
-		config.userChannels = null;
-		config.allowedTxPowers = null;
-		config.autoTxPowers = null;
-		config.userTxPowers = null;
```
pohanhf, line 90: We need to keep them as the optimization algorithms use these default values as the conditions to throw an exception or trigger some certain logic. :)

RockyMandayam2: Java automatically initializes uninitialized Object member variables to null, so having these does not add literally - but Jeffrey told me to keep these here anyways just for extra clarity for developers.

PR #16, opened 2022-07-27
[WIFI-10387] Adapt rrm service configuration for successful integration into Docker Compose deployment

oblom0v:
Signed-off-by: Johann Hoffmann <johann.hoffmann@mailbox.org>

### DIFF
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -57,6 +57,15 @@ ingresses:
       serviceName: owrrm
       servicePort: restapi
 
+volumes:
+  owrrm:
+    # Change this if you want to use another volume type
+    - name: persist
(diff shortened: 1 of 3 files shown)


helm/values.yaml
```diff
@@ -77,14 +86,19 @@ affinity: {}
 
 podAnnotations: {}
 
+persistence:
+  enabled: true
+  # storageClassName: "-"
+  accessModes:
+    - ReadWriteOnce
+  size: 1Gi
```
elludraon, line 94: These are tiny JSON files, 100MB would be more than enough but up to you.

elludraon (approved):

PR #17, opened 2022-07-27
Aggregate wifi scan entries

RockyMandayam2:


src/main/java/com/facebook/openwifirrm/modules/Modeler.java
```diff
@@ -340,6 +343,57 @@ private void processData(InputData data) {
 		}
 	}
 
+	/**
+	 * NOTE: if a BSSID does not have a non-obsolete entry, it will be returned
+	 * (i.e., it will not be a key in the returned map).
+	 */
+	public Map<String, WifiScanEntry> getAggregatedWifiScans(long obsoletion_period) {
```
elludraon, line 350: Input variable name should include units (ex. `obsoletionPeriodMs`) and you should probably also mention it in the Javadoc comment

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
```diff
@@ -340,6 +343,57 @@ private void processData(InputData data) {
...
+		Map<String, WifiScanEntry> aggregatedWifiScans = new HashMap<>();
+		for (String bssid : dataModel.latestWifiScans.keySet()) {
+			// flatten the wifiscan entries and sort in reverse chronological order
+			List<List<WifiScanEntry>> scans = dataModel.latestWifiScans.get(bssid);
```
elludraon, line 354: Iterate on `map.entrySet()` so that you can call `entry.value()` instead of doing an unneeded lookup

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
```diff
@@ -340,6 +343,57 @@ private void processData(InputData data) {
...
+			List<WifiScanEntry> mostRecentToOldest = scans.stream().flatMap(list -> list.stream())
+					.sorted((entry1, entry2) -> {
+						return -Long.signum(entry1.tsf - entry2.tsf);
```
elludraon, line 357: I've never seen "signum" used before... I think you just want `Long.compare(a, b)`?

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
```diff
@@ -340,6 +343,57 @@ private void processData(InputData data) {
...
+					}).collect(Collectors.toUnmodifiableList());
+
+			/*
+			 * For a given BSSID, discard the obsolete entries. Then,consider the most
+			 * recent entry. Take its ht_oper and vht_oper. Discard earlier entries with a
+			 * different ht_oper or a different vht_oper. For earlier entries with the same
+			 * ht_oper and vht_oper, take the average signal value.
+			 */
+			long now = Instant.now().getEpochSecond();
```
elludraon, line 366: I think you want higher resolution here, e.g. for milliseconds use `System.currentTimeMillis()`. Also, typically `now` should be defined outside the loop unless there is a good reason not to.

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
```diff
@@ -340,6 +343,57 @@ private void processData(InputData data) {
...
+			String newest_ht_oper = null;
+			String newest_vht_oper = null;
+			double averagedSignal = 0.0;
+			int count = 0;
+			for (WifiScanEntry entry : mostRecentToOldest) {
+				if (now - entry.tsf >= obsoletion_period) {
```
elludraon, line 372: `now` and `tsf` are probably not in the same time reference, we probably just need to record the overall scan response timestamp in system time

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
```diff
@@ -340,6 +343,57 @@ private void processData(InputData data) {
...
+					averagedSignal = entry.signal;
+					count++;
+					continue;
+				}
+				if (!entry.ht_oper.equals(newest_ht_oper) || !entry.vht_oper.equals(newest_vht_oper)) {
+					// discard older entries with different ht_oper or different vht_oper
+					continue;
+				}
+				// average signal value from older entries with the same ht_oper and vht_oper
+				averagedSignal = (count / (count + 1)) * averagedSignal + (entry.signal / (count + 1));
```
elludraon, line 390: I'd add an input argument for this, not all use cases will want an average signal (say, maybe we want min+max). We can figure this out later as needed.

Also I'm pretty sure this is totally broken because you haven't casted "count" to floating point, so `count / (count + 1)` will always evaluate to 0.

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
```diff
@@ -340,6 +343,57 @@ private void processData(InputData data) {
...
+			for (WifiScanEntry entry : mostRecentToOldest) {
+				if (now - entry.tsf >= obsoletion_period) {
+					// discard obsolete entries
+					break;
+				}
+				if (newest_ht_oper == null || newest_vht_oper == null) {
+					// start with the most recent entry
+					newest_ht_oper = entry.ht_oper;
+					newest_vht_oper = entry.vht_oper;
+					aggregatedWifiScans.put(bssid, entry);
```
elludraon, line 380: I'm pretty sure you might see only one of `ht_oper` or `vht_oper` set, so this line is going to clobber any previous entry

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
paulvarkey: if `Long.compare(a, b)` works, then, it might already be the default behavior of `sorted`?

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
```diff
@@ -340,6 +343,57 @@ private void processData(InputData data) {
...
+					continue;
+				}
+				if (!entry.ht_oper.equals(newest_ht_oper) || !entry.vht_oper.equals(newest_vht_oper)) {
+					// discard older entries with different ht_oper or different vht_oper
+					continue;
+				}
+				// average signal value from older entries with the same ht_oper and vht_oper
+				averagedSignal = (count / (count + 1)) * averagedSignal + (entry.signal / (count + 1));
+			}
+			aggregatedWifiScans.get(bssid).signal = (int) Math.round(averagedSignal);
```
paulvarkey, line 392: surprised that there is only integer precision for the signal

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
paulvarkey, line 390: should `count` be incremented after this line?

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
pohanhf: ht_oper or vht_oper might be NULL in data so we need to take care of that case as well. For example, for 2.4G radio in HT mode, you might see vht_oper is NULL. We also need to aggregate the entires even though vht_oper is NULL for all the entries.

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
RockyMandayam2: Yes.

RockyMandayam2:
I will look into the failed checks soon, but not there yet

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
RockyMandayam2: The signal is measured in dBm. Po-Han mentioned that taking a raw mean (mean = sum/count) is fine for this use case even though this would return a mean of log-values (dBm) instead of the mean of the "original" (not log) values.

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
RockyMandayam2: Now should be defined outside the loop, yes. I'm still looking into confirming whether there is some reference/start for tsf. I'm almost certain the units are microseconds, but I don't know when "time = 0" is. As Jeffrey says in a different comment, we may have to store the time ourselves.

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
RockyMandayam2: Still looking into this.

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
pohanhf: We don't need to know time = 0 actually. We just need to compare the latest tsf and the tsf of the other entries. Would you mind pointing me to where @elludraon said to store the time ourselves? Thanks.

pohanhf (changes requested):

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
```diff
@@ -41,36 +42,33 @@ public class UCentralUtils {
...
-		public JsonArray ies;
+	/**
+	 * Wraps {@code WifiScanEntry} to include the time, because it is unclear if the
+	 * {@code tsf} field in {@code WifiScanEntry} is always present and is relative
+	 * to a constant reference time
+	 *
+	 * @author rockymandayam
+	 *
+	 */
+	public static class WifiScanEntryWrapper {
```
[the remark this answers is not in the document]
RockyMandayam2: I think I agree, I will make that change.

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -202,4 +208,70 @@ public static double calculateTPCMetrics(
...
+	 *
+	 * @param dataModel
+	 * @param obsoletionPeriodMs the maximum amount of time (in milliseconds) it is
+	 *                           worth aggregating over starting from the most
+	 *                           recent scan entry and working backwards in time
+	 * @param agg                an aggregator to output a calculated "aggregated
+	 *                           signal strength"
+	 * @return
+	 */
+	public Map<String, WifiScanEntryWrapper> getAggregatedWifiScans(Modeler.DataModel dataModel, long obsoletionPeriodMs,
```
pohanhf, line 222: I think overall this function is great :). Thanks for improving it based on the suggestions. Just leave a small comment here for you to consider. 

1. I think latestWifiScans is sorted chronologically already (in a reverse order). 
2. Even if latestWifiScans was not sorted, I feel like we might not need to do sorting here. The current method includes the sorting (O(NlogN)) and the iteration (O(N) with an early stop so still the total complexity is O(NlogN)). If we don't do the sorting at all, the complexity is only O(N). Since we only save 10-20 scan results, the performance might not have any difference. Just FYI. 

Also, I would expect we will have some unit tests to cover this function. :) Thanks.

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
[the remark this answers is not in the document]
pohanhf: Thanks. :)

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
RockyMandayam2: latestWifiScans is sorted in that items are added as they are received. I wasn't sure that meant chronologically. E.g., what if (due to network issue, concurrency, anything) an earlier scan entry arrived later at the rrm code? Is that possible?

RockyMandayam2: Yes, I'm still working on writing unit tests...

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
```diff
@@ -45,15 +45,51 @@ private UCentralUtils() {}
...
+		/** Signal strength measured in dBm */
 		public int signal;
+		/** BSSID is the MAC address of the device */
 		public String bssid;
 		public String ssid;
 		public long tsf;
+		/**
+		 * ht_oper is short for "high throughput operator". This field contains some
+		 * information already present in other fields. This is because this field was
+		 * added later in order to capture the non-redundant information but also
```
paulvarkey, line 57:
```suggestion
		 * added later in order to capture some new information but also
```

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
```diff
@@ -45,15 +45,51 @@ private UCentralUtils() {}
...
+		 * includes some redundant information. 802.11 defines the HT operator and
+		 * vendors may define additional fields.
+		 *
+		 * This field is specified as 24 bytes, but it is encoded in base64. 24 bytes is
+		 * 192 bits which is 32 base64 characters, so this field would be a 32-character
+		 * string. However, I believe the first byte (the Element ID, which should be 61
```
paulvarkey, line 63:
```suggestion
		 * string. However, it is likely the case that the first byte (the Element ID, which should be 61
```

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
```diff
@@ -45,15 +45,51 @@ private UCentralUtils() {}
...
+		 * for ht_oper) and the second byte (Length) are omitted when the wifi scan
+		 * results are sent to rrm. Typically in base64, the input has bytes padded
+		 * until its length is divisible by 6, so the 22 bytes would be padded by two
+		 * bytes to reach 24 bytes. Typically, the ASCII value of "=" is padded, which
+		 * would explain why ht_oper values always end with two equals signs.
```
paulvarkey, line 68:
```suggestion
		 * explains why ht_oper values always end with two equals signs.
```

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
```diff
@@ -45,15 +45,51 @@ private UCentralUtils() {}
...
+		 */
 		public String ht_oper;
+		/**
+		 * vht_oper is short for "very high throughput operator". This field contains
+		 * some information already present in other fields. This is because this field
+		 * was added later in order to capture the non-redundant information but also
```
paulvarkey, line 74:
```suggestion
		 * was added later in order to capture some new information but also
```

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
```diff
@@ -45,15 +45,51 @@ private UCentralUtils() {}
...
+		 * includes some redundant information. 802.11 defines the HT operator and
```
paulvarkey, line 75:
```suggestion
		 * includes some redundant information. 802.11 defines the VHT operator and
```

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
```diff
@@ -45,15 +45,51 @@ private UCentralUtils() {}
...
+		 * please see the javadoc for ht_oper.
+		 */
 		public String vht_oper;
 		public int capability;
 		public int frequency;
+		/** IE = information element */
 		public JsonArray ies;
+		/**
+		 * Unix time in milliseconds (ms). This field is not defined in the uCentral
+		 * API. We added it because it was unclear whether the {@code tsf} field was a
```
paulvarkey, line 88:
```suggestion
		 * API. This is added because it is unclear whether the {@code tsf} field is a
```

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
```diff
@@ -45,15 +45,51 @@ private UCentralUtils() {}
...
+		 * reliable time measurement (e.g., we did not know if/when it got reset or if
```
paulvarkey, line 89:
```suggestion
		 * reliable time measurement (e.g., it is unclear if/when it gets reset or if
```

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
```diff
@@ -45,15 +45,51 @@ private UCentralUtils() {}
...
+		 * it was always available).
```
paulvarkey, line 90:
```suggestion
		 * it is always available).
```

src/test/java/com/facebook/openwifirrm/modules/aggregators/MeanAggregatorTest.java
```diff
@@ -0,0 +1,45 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInfo;
+
+public class MeanAggregatorTest {
+
+	private static double eps;
```
paulvarkey, line 11: do you envision many more tests in this class, that would need a global `eps`?

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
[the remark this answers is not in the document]
RockyMandayam2: Jeffrey said to keep the interface model "clean". For one, we may want to import automatically from the yamls later, which would be more difficult if we had our own fields added to these models. So I will either use a wrapper or extend the `WifiScanEntry` class. (This is all mostly "implementation details" and doesn't change the fact that we want to store our own timestamp)

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -202,4 +209,71 @@ public static double calculateTPCMetrics(
 			return Double.POSITIVE_INFINITY;
 		}
 	}
+
+	/**
+	 *
+	 * @param dataModel
+	 * @param obsoletionPeriodMs the maximum amount of time (in milliseconds) it is
+	 *                           worth aggregating over starting from the most
```
paulvarkey, line 217:
```suggestion
	 *                           worth aggregating over, starting from the most
```

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -202,4 +209,71 @@ public static double calculateTPCMetrics(
...
+	 * @param agg                an aggregator to output a calculated "aggregated
+	 *                           signal strength"
+	 * @return a map from BSSID to an "aggregated wifiscan entry" (i.e., a single
+	 *         entry with its {@code signal} attribute modified to be the aggregated
+	 *         signal value instead of the value in just the most recent entry.
+	 */
+	public Map<String, WifiScanEntry> getAggregatedWifiScans(Modeler.DataModel dataModel, long obsoletionPeriodMs,
+			Aggregator<Double> agg) {
+		/*
+		 * NOTE: if a BSSID does not have a non-obsolete entry, it will be returned
```
paulvarkey, line 228:
```suggestion
		 * NOTE: if a BSSID does not have a non-obsolete entry, it will not be returned
```

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -202,4 +209,71 @@ public static double calculateTPCMetrics(
...
+					newestVhtOper = entry.vht_oper;
+				}
+				// if the entry matches ht_oper and vht_oper, add its signal to the aggregate
+				if ((entry.ht_oper == null || entry.ht_oper == newestHtOper)
+						&& (entry.vht_oper == null || entry.vht_oper == newestVhtOper)) {
+					aggregatedWifiScans.putIfAbsent(bssid, entry);
+					agg.addValue((double) entry.signal);
+				}
+			}
+			aggregatedWifiScans.get(bssid).signal = (int) Math.round(agg.getAggregate());
```
paulvarkey, line 275: will this aggregate across `bssid`s -- I thought that each one needed to be separately aggregated. Do we need to reset agg somewhere?

paulvarkey (changes requested):

src/test/java/com/facebook/openwifirrm/modules/aggregators/MeanAggregatorTest.java
RockyMandayam2: No, actually I can move it into the method. It can be moved back out later (perhaps to even a parent class) if it turns out there are other aggregators and they all want to share the same `eps`

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
RockyMandayam2: Good catch.

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -202,4 +208,73 @@ public static double calculateTPCMetrics(
...
+					newestHtOper = entry.ht_oper;
+				}
+				if (newestVhtOper == null) {
+					// may still be null after this assignment
+					newestVhtOper = entry.vht_oper;
+				}
+				// if the entry matches ht_oper and vht_oper, add its signal to the aggregate
+				if ((entry.ht_oper == null || entry.ht_oper == newestHtOper)
+						&& (entry.vht_oper == null || entry.vht_oper == newestVhtOper)) {
+					aggregatedWifiScans.putIfAbsent(bssid, new ProcessedWifiScanEntry(entry));
```
paulvarkey, line 272: do we need to add a copy here, as opposed to merely placing the entry directly?

src/main/java/com/facebook/openwifirrm/mysql/DatabaseManager.java
```diff
@@ -482,6 +482,8 @@ public void addWifiScan(
 				stmt.setLong(4, entry.last_seen);
 				stmt.setInt(5, entry.signal);
 				stmt.setInt(6, entry.channel);
+				// TODO is this right? How do I test this?
```
paulvarkey, line 485:
```suggestion
				// TODO : untested
```

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
paulvarkey: bump

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
paulvarkey: bump

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
```diff
@@ -42,35 +44,55 @@ public class UCentralUtils {
...
+				return false;
+			}
+			if (getClass() != obj.getClass()) {
+				return false;
+			}
+			ProcessedWifiScanEntry other = (ProcessedWifiScanEntry) obj;
+			return unixTimeMs == other.unixTimeMs;
+		}
+
+		// TODO should I NOT implement hashCode since this is mutable, or I SHOULD
```
paulvarkey, line 82: delete comment -- overriding `hashCode` is the best practice

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
```diff
@@ -42,35 +44,55 @@ public class UCentralUtils {
...
+			final int prime = 31;
+			int result = super.hashCode();
+			result = prime * result + Objects.hash(unixTimeMs);
+			return result;
+		}
+
+		@Override
+		public String toString() {
+			return String.format("ProcessedWifiScanEntry[signal=%d, bssid=%s, unixTimeMs=%d]", signal, bssid,
+					unixTimeMs);
```
paulvarkey, line 95: optional : format `unixTimeMs` into human-readable string (https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html)

src/main/java/com/facebook/openwifirrm/ucentral/models/WifiScanEntry.java
```diff
@@ -0,0 +1,100 @@
...
+			return false;
+		}
+		WifiScanEntry other = (WifiScanEntry) obj;
+		return Objects.equals(bssid, other.bssid) && capability == other.capability && channel == other.channel
+				&& frequency == other.frequency && Objects.equals(ht_oper, other.ht_oper)
+				&& Objects.equals(ies, other.ies) && last_seen == other.last_seen && signal == other.signal
+				&& Objects.equals(ssid, other.ssid) && tsf == other.tsf && Objects.equals(vht_oper, other.vht_oper);
+	}
+
+	// TODO should I NOT implement hashCode since this is mutable, or I SHOULD
```
paulvarkey, line 93: delete comment -- implementing hashCode is good practice

src/test/java/com/facebook/openwifirrm/modules/ModelerUtilsTest.java
```diff
@@ -77,4 +86,61 @@ void testValidMetric() throws Exception {
...
+		aggregatedEntryB = ModelerUtils.getAggregatedWifiScans(dataModel, obsoletionPeriodMs, new MeanAggregator())
+				.get(bssidB);
+		ProcessedWifiScanEntry expectedAggregatedEntry = new ProcessedWifiScanEntry(entryB2);
+		expectedAggregatedEntry.signal = -61; // average of -60 and -62
+		assertFalse(dataModel.latestWifiScans.containsKey(bssidA));
+		assertEquals(expectedAggregatedEntry, aggregatedEntryB);
+
+		// test the obsoletion period
+		signal = -64;
+		timeOffsetMs = obsoletionPeriodMs; // now only entry 2 and entry 3 should be aggregated
```
paulvarkey, line 133: it's not clear to me why entry 2 (60 s) is in the obsoletion period and entry 1 (0 s) is not, if the period is 900 s.

paulvarkey (changes requested):

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
RockyMandayam2: Yes, because in the aggregate entry, the `signal` value gets modified to reflect the average. If I put the entry directly, I'd modify that entry, which would modify the same object which is in `latestWifiScans` (which is supposed to be raw non-aggregated entries)

src/test/java/com/facebook/openwifirrm/modules/ModelerUtilsTest.java
RockyMandayam2: The obsoletion period is "exclusive" in that if the obsoletion period is X ms, then a record X ms ago is considered obsolete. That was just an arbitrary choice made by me (though I think it does avoid the need for a +/- 1 somewhere in the code so it is easier). I will make a clarifying comment.

pohanhf:
Is it possible that we don’t touch the functions in the optimizer in this PR? Could we let still those functions use wifiScanEntry? The following are my concerns:
1. We might consider to use both latest wifiscanentry and aggregated wifiscanentry later on. Let’s get the aggregated one ready in this PR, and then work on the design choice later. (Also, once the aggregation is done, we don’t need unixTimeMs anymore. Therefore, we can always fall back to use the parent class. Just FYI.)
2. Unless we could guarantee all these changes working very well in real setups, let’s separate it out so it is easier for us to debug. 

Please let me know if you have any other concerns. Thanks. Cc @elludraon @paulvarkey

RockyMandayam2:
> Thanks for improving the PR and refactoring many places in the existing code. :) Is it possible that we don’t touch the functions in the optimizer in this PR? Could we let those functions still use wifiScanEntry? The following are my concerns:
> 
> 1. We might consider to use both latest wifiscanentry and aggregated wifiscanentry later on. Let’s get the aggregated one ready in this PR, and then work on the design choice later. (Also, once the aggregation is done, we don’t need unixTimeMs anymore. Therefore, we can always fall back to use the parent class. Just FYI.)
> 2. Unless we could guarantee all these changes working very well in real setups, let’s separate it out so it is easier for us to debug.
> 
> Please let me know if you have any other concerns. Thanks. Cc @elludraon @paulvarkey

I agree about splitting the PR into multiple, I can do that. @elludraon what are your thoughts about using the old class vs the new class?

RockyMandayam2:
> > Thanks for improving the PR and refactoring many places in the existing code. :) Is it possible that we don’t touch the functions in the optimizer in this PR? Could we let those functions still use wifiScanEntry? The following are my concerns:
> > 
> > 1. We might consider to use both latest wifiscanentry and aggregated wifiscanentry later on. Let’s get the aggregated one ready in this PR, and then work on the design choice later. (Also, once the aggregation is done, we don’t need unixTimeMs anymore. Therefore, we can always fall back to use the parent class. Just FYI.)
> > 2. Unless we could guarantee all these changes working very well in real setups, let’s separate it out so it is easier for us to debug.
> > 
> > Please let me know if you have any other concerns. Thanks. Cc @elludraon @paulvarkey
> 
> I agree about splitting the PR into multiple, I can do that. @elludraon what are your thoughts about using the old class vs the new class?

I think maybe I could split it into 5 PRs (maybe the last two could be combined):
- combining literals and duplicate variables into constants
- documentation (code comments)
- creating those util methods to more easily generate ht_oper and vht_oper and the corresponding tests
- stuff relating to creating ProcessedWifiScanEntry
- implement the aggregating method and the corresponding tests

Thoughts? @elludraon @paulvarkey

pohanhf:
> > > Thanks for improving the PR and refactoring many places in the existing code. :) Is it possible that we don’t touch the functions in the optimizer in this PR? Could we let those functions still use wifiScanEntry? The following are my concerns:
> > > 
> > > 1. We might consider to use both latest wifiscanentry and aggregated wifiscanentry later on. Let’s get the aggregated one ready in this PR, and then work on the design choice later. (Also, once the aggregation is done, we don’t need unixTimeMs anymore. Therefore, we can always fall back to use the parent class. Just FYI.)
> > > 2. Unless we could guarantee all these changes working very well in real setups, let’s separate it out so it is easier for us to debug.
> > > 
> > > Please let me know if you have any other concerns. Thanks. Cc @elludraon @paulvarkey
> > 
> > 
> > I agree about splitting the PR into multiple, I can do that. @elludraon what are your thoughts about using the old class vs the new class?
> 
> I think maybe I could split it into 5 PRs (maybe the last two could be combined):
> 
> * combining literals and duplicate variables into constants
> * documentation (code comments)
> * creating those util methods to more easily generate ht_oper and vht_oper and the corresponding tests
> * stuff relating to creating ProcessedWifiScanEntry
> * implement the aggregating method and the corresponding tests
> 
> Thoughts? @elludraon @paulvarkey

The list looks good. I think the first two can be one as they are related to code refactoring. The following three can be another one as they are more related to the wifiscan aggregation. 

If you think using the new class makes more sense to you, you could just rename the parent class to something else (e.g., rawWifiScanEntry) and then let the child class use wifiScanEntry so you don’t touch the interface between Optimizer and Modeler (and you don’t modify too many files all at once). Thanks.

RockyMandayam2:
> > > > Thanks for improving the PR and refactoring many places in the existing code. :) Is it possible that we don’t touch the functions in the optimizer in this PR? Could we let those functions still use wifiScanEntry? The following are my concerns:
> > > > 
> > > > 1. We might consider to use both latest wifiscanentry and aggregated wifiscanentry later on. Let’s get the aggregated one ready in this PR, and then work on the design choice later. (Also, once the aggregation is done, we don’t need unixTimeMs anymore. Therefore, we can always fall back to use the parent class. Just FYI.)
> > > > 2. Unless we could guarantee all these changes working very well in real setups, let’s separate it out so it is easier for us to debug.
> > > > 
> > > > Please let me know if you have any other concerns. Thanks. Cc @elludraon @paulvarkey
> > > 
> > > 
> > > I agree about splitting the PR into multiple, I can do that. @elludraon what are your thoughts about using the old class vs the new class?
> > 
> > 
> > I think maybe I could split it into 5 PRs (maybe the last two could be combined):
> > 
> > * combining literals and duplicate variables into constants
> > * documentation (code comments)
> > * creating those util methods to more easily generate ht_oper and vht_oper and the corresponding tests
> > * stuff relating to creating ProcessedWifiScanEntry
> > * implement the aggregating method and the corresponding tests
> > 
> > Thoughts? @elludraon @paulvarkey
> 
> The list looks good. I think the first two can be one as they are related to code refactoring. The following three can be another one as they are more related to the wifiscan aggregation.
> 
> If you think using the new class makes more sense to you, you could just rename the parent class to something else (e.g., rawWifiScanEntry) and then let the child class use wifiScanEntry so you don’t touch the interface between Optimizer and Modeler (and you don’t modify too many files all at once). Thanks.

Re: doing 2 PRs, that sounds good.
Re: renaming the parent class - we should not do that because the name (and class) is defined by the open wifi team, and Jeffrey also mentioned he wanted to keep that interface clean... We'll wait to see his thoughts on this as well.

RockyMandayam2:
> 

Jeffrey mentioned to me (and I agree) that using WifiScanEntry in some places and ProcessedWifiScanEntry in others may be difficult (e.g., it is hard to cast between List<ChildClass> and List<ParentClass>. As I mentioned in my previous comment, we also don't want to rename the parent class.

Also, as for not needing unixTimeMS after the aggregation - I think it could still be useful, e.g., for some kind of diagnosis or root cause analysis later.

So all in all, I will keep the current structure, unless you have more concerns Po-Han :) (or perhaps I misunderstood you).

pohanhf:
> > 
> 
> Jeffrey mentioned to me (and I agree) that using WifiScanEntry in some places and ProcessedWifiScanEntry in others may be difficult (e.g., it is hard to cast between List and List. As I mentioned in my previous comment, we also don't want to rename the parent class.
> 
> Also, as for not needing unixTimeMS after the aggregation - I think it could still be useful, e.g., for some kind of diagnosis or root cause analysis later.
> 
> So all in all, I will keep the current structure, unless you have more concerns Po-Han :) (or perhaps I misunderstood you).

No, I believe the structure is defined by the OpenWiFi team but not the name. The class name is flexible (I might be wrong :) ). If everyone agrees with the change, then I am okay with that.

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -202,4 +208,73 @@ public static double calculateTPCMetrics(
...
+					agg.addValue((double) entry.signal);
+				}
+			}
```
pohanhf, line 275: I got some time to read through this. :) I think one case might be missing in the current implementation. 

If the most recent wifiscan has ht_oper = null and vht_oper = null, then you need to check the second most recent one to see whether it is still ht_oper = null and vht_oper = null. The reason is because its firmware doesn't support the wifiscan feature but we still need to aggregate them. In the current logic, it will skip this case in line 259. 

I believe the easiest way to do this is to assign newestHtOper =  mostRecentToOldest.get(0).ht_oper and newestVhtOper = mostRecentToOldest.get(0).vht_oper and then check the delta between the most recent one and the remaining entires. 

Again, I might be wrong :). Just share what I think. Please ignore my comment if you think it doesn't make sense. I will remove the change request in this PR so you won't be blocked. Thanks. :)

RockyMandayam2: Interesting, I did not realize this. What does it mean for the firmware to not support the scan? If that were the case, why would we have a `WifiScanEntry` at all? And in this case, does it make sense to aggregate, if we don't know the channel info? (I'm assuming if it doesn't support a scan, the other fields would be invalid as well)

pohanhf: Oh sorry that I should have been more precise. “The wifiscan feature” here means a valid wifiscan but not reporting those two fields (ht_oper and vht_oper). In this case, we still have the rest of things (including channel, signal, etc.,) but not those two. This happens when someone use the older (but still valid) version of firmware, i.e., <2.4 (those two fields are supported after 2.5). 

Again, I could be wrong :) so please feel free to skip this comment. Just share what I have seen in the data before.

RockyMandayam2: Got it. What you're saying makes sense. In that case, would we want to do some kind of checking before aggregating still? E.g., check that `channel` and `frequency` are the same and aggregate only entries with the same channel and frequency?

pohanhf: Those two are good IMHO. Thanks. :)

src/main/java/com/facebook/openwifirrm/mysql/DatabaseManager.java
RockyMandayam2: Will change in the new PR (since we're splitting this PR into two).

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
RockyMandayam2: Will change in the new PR, thank you!

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
RockyMandayam2: Will change in the new PR, thank you!

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
RockyMandayam2: Will change in the new PR, thank you!

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
RockyMandayam2: I think I will leave this as is for now since I use it more for debugging for now, where the raw value is easier to use :)

src/main/java/com/facebook/openwifirrm/ucentral/models/WifiScanEntry.java
RockyMandayam2: Will change in the new PR, thank you!

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
RockyMandayam2: I'll work on this in the separate PR, thank you!

src/test/java/com/facebook/openwifirrm/modules/ModelerUtilsTest.java
RockyMandayam2: This will be handled in the new PR.

RockyMandayam2:
I am closing this PR due to the fact that I am splitting this into two separate PRs. One of those is already merged in, and one of them is coming soon.

PR #18, opened 2022-07-29
implement validateToken and wire it up in ApiServer

jwoos:
Implement `validateToken` in `uCentralClient` and call from ApiServer to make sure that the request is valid. 

Generated a token, then tested

```
$ curl -k -H "Accept: application/json" -H "Authorization: Bearer SOME_GOOD_TOKEN" -i 'localhost:16789/api/v1/getTopology'
HTTP/1.1 200 OK
Date: Thu, 28 Jul 2022 21:42:08 GMT
Server:
Content-Type: application/json
Transfer-Encoding: chunked

# resp

$ curl -k -H "Accept: application/json" -H "Authorization: Bearer SOME_BAD_TOKEN" -i 'localhost:16789/api/v1/getTopology'
HTTP/1.1 403 Forbidden
Date: Thu, 28 Jul 2022 21:42:14 GMT
Server:
Content-Type: text/html;charset=utf-8
Transfer-Encoding: chunked
```

### DIFF
--- a/src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
+++ b/src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
@@ -8,10 +8,12 @@
 
 package com.facebook.openwifirrm.ucentral;
 
+import java.time.Instant;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ThreadLocalRandom;
 
 import org.json.JSONArray;
@@ -20,7 +22,6 @@ import org.json.JSONObject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.facebook.openwifirrm.RRMConfig.UCentralConfig.UCentralSocketParams;
 import com.facebook.openwifirrm.ucentral.gw.models.CommandInfo;
 import com.facebook.openwifirrm.ucentral.gw.models.DeviceCapabilities;
 import com.facebook.openwifirrm.ucentral.gw.models.DeviceConfigureRequest;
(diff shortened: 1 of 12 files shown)


src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
```diff
@@ -614,6 +624,39 @@ public void setServiceEndpoint(String service, ServiceEvent event) {
 		}
 	}
 
+	/**
+	 * Validate the token is still valid.
+	 */
+	public boolean validateToken(String token) {
+		if (!tokenCache.containsKey(token)) {
```
elludraon, line 631: nit: just to avoid the double-lookup (with line 655), do:
```
Long expiry = tokenCache.get(token);
if (expiry == null) {
  // ...
  expiry = result.tokenInfo.created + result.tokenInfo.expires_in;
  tokenCache.put(token, expiry);
}
```

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
```diff
@@ -614,6 +624,39 @@ public void setServiceEndpoint(String service, ServiceEvent event) {
...
+						"Failed to deserialize to TokenValidationResult: %s", response.getBody()
+				);
+				logger.error(errMsg, e);
+				return false;
+			}
+
+			tokenCache.put(token, result.tokenInfo.created + result.tokenInfo.expires_in);
+		}
+
+		long expiry = tokenCache.get(token);
```
elludraon, line 655: this will be covered by the above comment, but note that this might crash in async/concurrent edge case if the "token" no longer exists (ex. got removed from cache somewhere else) and "null" can't auto-unbox into "long" (will raise NullPointerException).

src/main/java/com/facebook/openwifirrm/ucentral/gw/models/UserInfo.java
```diff
@@ -0,0 +1,45 @@
...
+    public String validationEmail;
+    public long validationDate;
+    public long created;
+    public String validationURI;
+    public long lastPasswordChange;
+    public long lastEmailCheck;
+    public long lastLogin;
+    public String currentLoginURI;
+    public String currentPassword;
+    public java.util.List<String> lastPasswords;
```
elludraon, line 29: You've already imported this, don't put the full package path here (`List` instead of `java.util.List`)

elludraon (commented):
Looks great! Please address a few comments inline.

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
```diff
@@ -8,10 +8,13 @@
 
 package com.facebook.openwifirrm.ucentral;
 
+import java.lang.Long;
```
elludraon, line 11: you don't need to import boxed types

elludraon (approved):

PR #19, opened 2022-07-30
[WIFI-9980] Add RRM scheduler

elludraon:

Allow scheduling RRM algorithms to be run per zone. This implementation uses the Quartz Job Scheduler library.
- Add "schedule" (`RRMSchedule`) field to `DeviceConfig`, which is only read at the network/zone layer (i.e. ignored at AP layer). This allows scheduling specified algorithms at a given interval in Quartz's cron-like format.
- Add `RRMScheduler` module encapsulating all Quartz implementation details and RRM execution.
- Add `RRMScheduler.syncTriggers()` calls throughout codebase as needed to ensure triggers remain in sync with topology and device config.
- Add "schedulerParams" (`RRMSchedulerParams`) structure to `RRMConfig`, which currently only allows specifying thread pool size.

**The config field names and values are TBD.** For now, this reuses the names from `ApiServer`.

Example `device_config.json` snippet which runs channel optimization with default settings every 5 minutes:
```json
{
  "zoneConfig": {
    "some-venue": {
      "schedule": {
        "cron": "0 0/5 * * * ?",
        "algorithms": [
          {
            "name": "optimizeChannel"
          }
        ]
      }
    }
  }
}
```

Unrelated:
- Delete commented optimizer code from `ProvMonitor` because it won't be invoked from there.
- Fix leakage of `DeviceConfig` references in `DeviceDataManager.computeDeviceConfig()` by returning a deep clone.
- Change RRM config field "configOnEventOnly" default from `false` to `true`.

### DIFF
--- a/src/main/java/com/facebook/openwifirrm/DeviceDataManager.java
+++ b/src/main/java/com/facebook/openwifirrm/DeviceDataManager.java
@@ -11,7 +11,9 @@ package com.facebook.openwifirrm;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.TreeMap;
@@ -295,6 +297,17 @@ public class DeviceDataManager {
 		return gson.fromJson(contents, DeviceTopology.class);
 	}
 
+	/** Return all zones in the topology. */
+	public List<String> getZones() {
+		Lock l = topologyLock.readLock();
+		l.lock();
+		try {
+			return new ArrayList<>(topology.keySet());
+		} finally {
+			l.unlock();
+		}
+	}
+
 	/** Return the RF zone for the given device, or null if not found. */
 	public String getDeviceZone(String serialNumber) {
 		if (serialNumber == null || serialNumber.isEmpty()) {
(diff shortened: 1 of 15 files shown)


src/main/java/com/facebook/openwifirrm/DeviceDataManager.java
```diff
@@ -539,4 +552,16 @@ public void updateDeviceApConfig(DeviceApConfigFunction fn) {
 		cachedDeviceConfigs.clear();
 		saveDeviceLayeredConfig();
 	}
+
+	/**
+	 * Return config for the given zone with all config layers applied. It is
+	 * the caller's responsibility to ensure the zone exists in the topology.
```
pohanhf, line 558: I guess the assumption here is that a zone (i.e., venue in owprov) will be created while (or before) RRM scheduler is being set, is this the correct understanding? Thanks.

pohanhf (approved):
Looks good to me. Thanks for implementing this important feature :). Would love to see the example snippet included in the doc as well (once the data format/interface is finalized).

src/main/java/com/facebook/openwifirrm/DeviceDataManager.java
elludraon: Sort of, not exactly.
- `ProvMonitor` calls `scheduler.syncTriggers()` after it syncs topology from owprov, so this is handled.
- We also want to decouple the scheduler from owprov such that the scheduler works even if owprov is not deployed.

PR #25, opened 2022-08-06
Aggregate wifiscans

RockyMandayam2:
The purpose of this PR is to enable aggregation of wifiscan entries so that optimizers have a more complete picture.

The `getAggregatedWifiScans()` method was added in `ModelerUtils` to do the core aggregation logic. To support this, other significant changes were made:
- An aggregators package to implement basic numerical aggregation methods used by `getAggregatedWifiScans()`
- Timestamps are needed for aggregation. The wifiscan entries have a `tsf` field but this is an unknown time reference. Therefore, this PR extends the base class with a subclass to keep track of the unix time (and the base class was moved to the ucentral.models package). This unix time is gotten from the Kafka records.
- The ucentral.operationelement package was added mainly to help decode the base64 encoded HT operation element and VHT operation element strings

Tests were added to test the added packages/classes and the `getAggregatedWifiScans()` in `ModelerUtils`


src/main/java/com/facebook/openwifirrm/ChannelWidth.java
```diff
@@ -0,0 +1,11 @@
+package com.facebook.openwifirrm;
+
+/**
+ * Represents the possible channel widths allowed in the 5 GHz band.
+ *
+ * @author rockymandayam
+ *
+ */
+public enum ChannelWidth {
```
RockyMandayam2, line 9: I think I should move this class - @elludraon any suggestions to where?

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -202,4 +209,89 @@ public static double calculateTPCMetrics(
...
+	 * therefore frequency, since they share a one-to-one mapping), that is enough
+	 * to aggregate them. However, 802.11n introduced channel bonding, so for
+	 * 802.11n onwards, channel width must be checked.
+	 *
+	 * @param entry1
+	 * @param entry2
+	 * @return true if the entries should be aggregated
+	 */
+	private static boolean matchesForAggregation(WifiScanEntry entry1, WifiScanEntry entry2) {
+		// TODO should we check the entire ht_oper and vht_oper or just channel width?
```
RockyMandayam2, line 228: I'd love feedback from the wireless systems experts here :)

src/main/java/com/facebook/openwifirrm/mysql/DatabaseManager.java
```diff
@@ -482,6 +482,8 @@ public void addWifiScan(
 				stmt.setLong(4, entry.last_seen);
 				stmt.setInt(5, entry.signal);
 				stmt.setInt(6, entry.channel);
+				// TODO: untested
+				stmt.setLong(7, entry.unixTimeMs);
```
RockyMandayam2, line 486: How do I test this and the other untested TODO in this file?

src/test/java/com/facebook/openwifirrm/modules/ModelerUtilsTest.java
```diff
@@ -77,4 +86,59 @@ void testValidMetric() throws Exception {
...
+		aggregatedEntryB = ModelerUtils.getAggregatedWifiScans(dataModel, obsoletionPeriodMs, new MeanAggregator())
+				.get(bssidB);
+		expectedAggregatedEntry = new WifiScanEntry(entryB3);
+		expectedAggregatedEntry.signal = -63; // average of -62 and -64
+		assertEquals(expectedAggregatedEntry, aggregatedEntryB);
+	}
+
+//	@Test
+//	void testMultipleEntriesAndAggregationLogic() {
+//		// TODO test 802.11n onwards (with non-null ht_oper and vht_oper)
```
RockyMandayam2, line 141: I will work on this for the next push (I just wanted to get a rough draft out since the aggregation logic has changed from my earlier PR)

src/test/java/com/facebook/openwifirrm/optimizers/TestUtils.java
```diff
@@ -65,11 +72,21 @@ public static JsonArray createDeviceStatusDualBand(int channel2G, int txPower2G,
 		return jsonList;
 	}
 
+	private static int channelToFrequency(int channel) {
```
RockyMandayam2, line 75: These two equations are actually taken from the IEEE 802.11 standard, I will add a comment mentioning that.

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -202,4 +209,89 @@ public static double calculateTPCMetrics(
...
+		return entry1.bssid.equals(entry2.bssid) && Objects.equals(entry1.ssid, entry2.ssid)
```
paulvarkey, line 229: is there a reason that three different styles are employed here for equality comparison?

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -202,4 +209,89 @@ public static double calculateTPCMetrics(
...
+			List<WifiScanEntry> mostRecentToOldest = scans.stream().flatMap(list -> list.stream())
+					.sorted((entry1, entry2) -> {
+						return -Long.compare(entry1.unixTimeMs, entry2.unixTimeMs);
+					}).collect(Collectors.toUnmodifiableList());
+
+			String bssid = mapEntry.getKey();
+			WifiScanEntry mostRecentEntry = mostRecentToOldest.get(0);
+			agg.reset();
+			for (WifiScanEntry entry : mostRecentToOldest) {
+				if (mostRecentEntry.unixTimeMs - entry.unixTimeMs >= obsoletionPeriodMs) {
```
paulvarkey, line 282: if I choose obsoletionPeriodMs as 0 (to indicate that I only want absolute recent), then, this loop will break and line 293 will be NullPointerException?

src/test/java/com/facebook/openwifirrm/optimizers/TestUtils.java
paulvarkey: Add a comment that the unit is MHz. Some folks like indicating that in the name itself -- e.g. `channelToFrequencyMHz`

paulvarkey, line 75: add some error checking here or at appropriate public call-sites for invalid channels -- perhaps, add list of valid channels to docstring

src/test/java/com/facebook/openwifirrm/optimizers/TestUtilsTest.java
```diff
@@ -0,0 +1,59 @@
+package com.facebook.openwifirrm.optimizers;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import org.junit.jupiter.api.Test;
+
+import com.facebook.openwifirrm.ChannelWidth;
+
+public class TestUtilsTest {
```
paulvarkey, line 9: this is a great test to keep this code stable -- feel free to add link to any public reference URL for the encoding logic

paulvarkey (changes requested):

halmehmood (commented):
Great work!

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
RockyMandayam2: For primitives, I do `==` (e.g., for an `int`). For Strings where it should never be null, I do `stringVariable.equals(otherObject)`. For Strings that could be `null` and where two null strings should count as being equal to each other, I do `Objects.equals`. Do you or @elludraon have a suggestion on a way to change this?

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
RockyMandayam2: Good point, perhaps making the non-obsolete window inclusive, meaning 0 ms => include the entry 0 ms ago. I'll try this.

src/test/java/com/facebook/openwifirrm/optimizers/TestUtilsTest.java
RockyMandayam2: There is no public reference URL, but I will add a docstring mentioning to look in the 802.11-2020 standard.

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
```diff
@@ -41,36 +43,56 @@ public class UCentralUtils {
...
+		@Override
+		public int hashCode() {
+			final int prime = 31;
+			int result = super.hashCode();
+			result = prime * result + Objects.hash(unixTimeMs);
+			return result;
+		}
+
+		@Override
+		public boolean equals(Object obj) {
```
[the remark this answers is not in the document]
RockyMandayam2: > Great work!

Thank you!

RockyMandayam2:
> Great work!

Thank you!

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
RockyMandayam2: Haleema said the basic MCS set does not need to be compared (it represents the intersection of the MCSs supported by all STAs in the network, which is not something that should affect whether we aggregate or not).

src/test/java/com/facebook/openwifirrm/modules/ModelerUtilsTest.java
```diff
@@ -77,4 +86,67 @@ void testValidMetric() throws Exception {
...
+		deviceDataManager.setTopology(TestUtils.createTopology("test-zone", apA, apB));
+		DataModel dataModel = new DataModel();
+
+		// if there are no scans (no list or empty list), there should be no aggregates
+		dataModel.latestWifiScans.computeIfAbsent(bssidB, k -> new LinkedList<>());
+		assertTrue(ModelerUtils.getAggregatedWifiScans(dataModel, obsoletionPeriodMs, new MeanAggregator()).isEmpty());
+
+		// for an AP, if there is one scan, the aggregate should just be that scan
+		// for the other device, there should be no aggregate
+		WifiScanEntry entryB1 = TestUtils.createWifiScanEntryWithBssid(1, bssidB); // defaults to signal = -60
```
elludraon, line 108: don't assume default values in another place, set anything here that you need

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -202,4 +211,109 @@ public static double calculateTPCMetrics(
...
+	private static boolean matchesForAggregation(WifiScanEntry entry1, WifiScanEntry entry2) {
+		// TODO should we check the entire ht_oper and vht_oper or just channel width?
+		return entry1.bssid.equals(entry2.bssid) && Objects.equals(entry1.ssid, entry2.ssid)
+				&& entry1.frequency == entry2.frequency && entry1.channel == entry2.channel
+				&& matchesHtForAggregation(entry1.ht_oper, entry2.ht_oper)
+				&& matchesVhtForAggregation(entry1.vht_oper, entry2.vht_oper);
+	}
+
+	/**
+	 *
```
elludraon, line 256: missing description?

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -202,4 +211,109 @@ public static double calculateTPCMetrics(
...
+	 * @param agg                an aggregator to output a calculated "aggregated
+	 *                           signal strength"
+	 * @return a map from BSSID to an "aggregated wifiscan entry" (i.e., a single
+	 *         entry with its {@code signal} attribute modified to be the aggregated
+	 *         signal value instead of the value in just the most recent entry.
+	 */
+	public static Map<String, WifiScanEntry> getAggregatedWifiScans(Modeler.DataModel dataModel,
+			long obsoletionPeriodMs,
+			Aggregator<Double> agg) {
+		assert obsoletionPeriodMs >= 0 : "obsoletionPeriodMs must be non-negative.";
```
elludraon, line 278: please don't ever `assert` in Java, throw `IllegalArgumentException` if needed

src/main/java/com/facebook/openwifirrm/modules/aggregators/Aggregator.java
```diff
@@ -0,0 +1,14 @@
+package com.facebook.openwifirrm.modules.aggregators;
```
elludraon, line 1: this should not be in `modules`

src/main/java/com/facebook/openwifirrm/modules/aggregators/Aggregator.java
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * Aggregates added values into one "aggregate" measure.
+ *
+ * @param <T>
```
elludraon, line 6: never put empty `@param`

src/main/java/com/facebook/openwifirrm/modules/operationelement/HTOperationElement.java
```diff
@@ -0,0 +1,162 @@
...
+	protected final boolean staChannelWidth;
+	protected final boolean rifsMode;
+	protected final byte htProtection;
+	protected final boolean nongreenfieldHtStasPresent;
+	protected final boolean obssNonHtStasPresent;
+	protected final byte channelCenterFrequencySegment2;
+	protected final boolean dualBeacon;
+	protected final boolean dualCtsProtection;
+	protected final boolean stbcBeacon;
+	protected final byte[] basicHtMcsSet;
```
elludraon, line 27: why are all of these `protected`? it seems like they should either be `public` or `private`

src/main/java/com/facebook/openwifirrm/modules/operationelement/HTOperationElement.java
```diff
@@ -0,0 +1,162 @@
...
+
+	/**
+	 * XXX some combinations of these parameters may be invalid as defined by
+	 * 802.11-2020, but this is not checked here. If fidelity to 802.11 is required,
+	 * the caller of this method must make sure to pass in valid parameters. The
+	 * 802.11-2020 specification has more details about the parameters.
```
elludraon, line 33: this note should be an inline comment, not a javadoc comment

src/main/java/com/facebook/openwifirrm/modules/operationelement/HTOperationElement.java
```diff
@@ -0,0 +1,162 @@
...
+	 *
+	 * @param primaryChannel                 channel index
```
elludraon, line 35: fix spacing

src/main/java/com/facebook/openwifirrm/modules/operationelement/HTOperationElement.java
```diff
@@ -0,0 +1,162 @@
...
+	 * @param secondaryChannelOffset
+	 * @param staChannelWidth
+	 * @param rifsMode
+	 * @param htProtection
+	 * @param nongreenfieldHtStasPresent
+	 * @param obssNonHtStasPresent
+	 * @param channelCenterFrequencySegment2
+	 * @param dualBeacon
+	 * @param dualCtsProtection
+	 * @param stbcBeacon
```
elludraon, line 45: never write empty `@param`

src/main/java/com/facebook/openwifirrm/modules/operationelement/HTOperationElement.java
```diff
@@ -0,0 +1,162 @@
...
+		this.channelCenterFrequencySegment2 = channelCenterFrequencySegment2;
+		this.dualBeacon = dualBeacon;
+		this.dualCtsProtection = dualCtsProtection;
+		this.stbcBeacon = stbcBeacon;
+		// the next 16 bytes are for the basic HT-MCS set
+		// a default is chosen; if needed, we can add a parameter to set these
+		this.basicHtMcsSet = new byte[16];
+	}
+
+	public HTOperationElement() {
```
elludraon, line 67: missing comment

src/main/java/com/facebook/openwifirrm/modules/operationelement/HTOperationElement.java
```diff
@@ -0,0 +1,162 @@
...
+		this((byte) 1, (byte) 0, false, false, (byte) 0, true, false, (byte) 0, false, false, false);
+	}
+
+	/**
+	 *
```
elludraon, line 72: missing comment

src/main/java/com/facebook/openwifirrm/modules/operationelement/OperationElement.java
```diff
@@ -0,0 +1,11 @@
+package com.facebook.openwifirrm.modules.operationelement;
```
elludraon, line 1: this should not be in "modules"

src/test/java/com/facebook/openwifirrm/modules/operationelement/VHTOperationElementTest.java
```diff
@@ -0,0 +1,51 @@
+package com.facebook.openwifirrm.modules.operationelement;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import org.junit.jupiter.api.Test;
+
+import com.facebook.openwifirrm.ChannelWidth;
+
+/** Test the Very High Throughput (VHT) Operation Element */
```
elludraon, line 9: you don't need a comment on these test classes, it's self-explanatory from the name

src/test/java/com/facebook/openwifirrm/optimizers/TestUtils.java
```diff
@@ -65,11 +69,30 @@ public static JsonArray createDeviceStatusDualBand(int channel2G, int txPower2G,
 		return jsonList;
 	}
 
+	/**
+	 *
```
elludraon, line 73: missing description

src/test/java/com/facebook/openwifirrm/optimizers/TestUtils.java
```diff
@@ -26,6 +28,8 @@ public class TestUtils {
 	/** The Gson instance. */
 	private static final Gson gson = new Gson();
 
+	public static final Instant DEFAULT_START_TIME = Instant.parse("2022-01-01T00:00:00Z");
```
elludraon, line 31: this needs a comment and the name is also not descriptive enough (ex. `DEFAULT_WIFISCAN_START_TIME`)

src/test/java/com/facebook/openwifirrm/modules/aggregators/MeanAggregatorTest.java
```diff
@@ -0,0 +1,41 @@
...
+		// adding 0 (the mean) does not change the mean
+		agg.addValue(0.0);
+		assertEquals(0, agg.getAggregate(), eps);
+
+		// add an "int"
+		agg.addValue(1.0);
+		assertEquals(0.5, agg.getAggregate(), eps);
+
+		// add a "float"
+		agg.addValue(3.5);
```
elludraon, line 26: this is not a float, it's a double (float literal needs `f` suffix)

src/main/java/com/facebook/openwifirrm/ChannelWidth.java
```diff
@@ -0,0 +1,8 @@
+package com.facebook.openwifirrm;
+
+/**
+ * Represents the possible channel widths allowed in the 5 GHz band.
+ */
+public enum ChannelWidth {
+	MHz_20, MHz_40, MHz_80, MHz_160, MHz_80_PLUS_80
+}
```
elludraon, line 8: I am not convinced this class is useful, and also the naming is bad (if this is specifically for 5GHz widths, that should be in the enum name)

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -202,4 +211,109 @@ public static double calculateTPCMetrics(
...
+	}
+
+	private static boolean matchesVhtForAggregation(String vhtOper1, String vhtOper2) {
+		if (vhtOper1 == vhtOper2) { // serves as a null check
+			return true;
+		}
+		VHTOperationElement vhtOperObj1 = new VHTOperationElement(vhtOper1);
+		VHTOperationElement vhtOperObj2 = new VHTOperationElement(vhtOper2);
+		return vhtOperObj1.matchesForAggregation(vhtOperObj2);
+	}
```
elludraon, line 231: these should just be static methods defined in the appropriate classes, not in `ModelerUtils`

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -202,4 +211,109 @@ public static double calculateTPCMetrics(
...
+			 * Flatten the wifiscan entries and sort in reverse chronological order. Sorting
+			 * is done just in case the entries in the original list are not chronological
+			 * already - although they are inserted chronologically, perhaps latency,
+			 * synchronization, etc. could cause the actual unixTimeMs to be out-of-order.
+			 */
+			List<List<WifiScanEntry>> scans = mapEntry.getValue();
+			if (scans.isEmpty()) {
+				continue;
+			}
+			List<WifiScanEntry> mostRecentToOldest = scans.stream().flatMap(list -> list.stream())
```
elludraon, line 295: nit: use `.flatMap(List::stream)` syntax when lambda is not necessary

src/main/java/com/facebook/openwifirrm/modules/operationelement/HTOperationElement.java
```diff
@@ -0,0 +1,162 @@
...
+	 * @param htOperString must be a String representing a base64 encoded properly
+	 *                     formatted ht operation element (see 802.11-2020 standard)
+	 */
+	public HTOperationElement(String htOperString) {
```
elludraon, line 76: NOTE: I have not checked this for correctness (or the VHT version). @pohanhf if you want to check when you're available.

elludraon (changes requested):

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
RockyMandayam2: I call the `matchesForAggregation` which is is in the appropriate class, but since a String is passed in here, I can't move this to the String class. I could just not use a method and instead just to the null check and object creation in the caller of this method. What do you think?

elludraon: I mean you can move these into `HTOperationElement` and `VHTOperationElement` and change to `public`

src/test/java/com/facebook/openwifirrm/modules/ModelerUtilsTest.java
RockyMandayam2: Hm, I think this may have consequences beyond just this. I definitely assume certain other things in this and other tests (like having/not having ht_oper, vht_oper, etc.). But I thought part of the purpose of this TestUtils functions are to provide a common/default way of doing things that are useful in tests? Should I try not to rely on things done by TestUtils methods?

src/main/java/com/facebook/openwifirrm/ChannelWidth.java
RockyMandayam2: It is not meant specifically for 5GHz widths. But in any case, I can remove it, since it is used in the operation element classes for encoding which you asked me to remove.

src/test/java/com/facebook/openwifirrm/modules/ModelerUtilsTest.java
elludraon: Sorry, I guess this is more of a specific comment for this call. If the test depends on a specific number, that input better be passed in as an argument, or you should reference the constant somewhere. Otherwise, if someone modifies the value in `TestUtils.createWifiScanEntryWithBssid()` it will be impossible to tell why this is breaking a seemingly unrelated test.

src/main/java/com/facebook/openwifirrm/modules/aggregators/Aggregator.java
RockyMandayam2: Any suggestions for the best place for this?

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
```diff
@@ -41,62 +43,55 @@ public class UCentralUtils {
...
-		 */
-		public String vht_oper;
-		public int capability;
-		public int frequency;
-		/** IE = information element */
-		public JsonArray ies;
+		public long unixTimeMs;
 
 		/** Default Constructor. */
 		public WifiScanEntry() {}
```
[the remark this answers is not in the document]
RockyMandayam2: super() is called by default, do you mean it's good to call it explicitly for style reasons?

src/main/java/com/facebook/openwifirrm/ChannelWidth.java
RockyMandayam2: He mentioned I can remove this class in another comment.

src/main/java/com/facebook/openwifirrm/mysql/DatabaseManager.java
RockyMandayam2: Just to clarify, what do you mean by "omit this"? I'm guessing you mean to keep the `setLong()` line but not include the `TODO` (since it will be consumed as part of another existing `TODO`), is that right?

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
RockyMandayam2: @elludraon any thoughts on this?

src/main/java/com/facebook/openwifirrm/mysql/DatabaseManager.java
elludraon: No, I mean to remove any changes to this file and deal with database cleanup later (this PR is already large enough).

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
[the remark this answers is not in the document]
elludraon: Oh I forgot. This is fine.

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
elludraon: I would also use `Objects.equals` for `bssid` but otherwise this is fine.

src/main/java/com/facebook/openwifirrm/modules/operationelement/HTOperationElement.java
RockyMandayam2: This is the correct spacing (using the style that all param descriptions should start to the right of ALL params, so in this case it starts to the far right because `channelCenterFrequencySegment2` is so long)

src/main/java/com/facebook/openwifirrm/modules/operationelement/OperationElement.java
RockyMandayam2: Any suggestions for the best place to put it?

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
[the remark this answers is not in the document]
RockyMandayam2: I already used the IDE-generated functions. I'l remove the comment.

src/main/java/com/facebook/openwifirrm/ucentral/models/WifiScanEntryResult.java
```diff
@@ -0,0 +1,100 @@
...
+	 * class is mutable!
+	 */
+	@Override
+	public int hashCode() {
+		return Objects.hash(bssid, capability, channel, frequency, ht_oper, ies, last_seen, signal, ssid, tsf,
+				vht_oper);
+	}
+
+	@Override
+	public boolean equals(Object obj) {
```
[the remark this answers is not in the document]
RockyMandayam2: Similar to the other instance of overriding equals, I'll just remove the comment on hashcode()

src/test/java/com/facebook/openwifirrm/modules/ModelerUtilsTest.java
```diff
@@ -77,4 +86,67 @@ void testValidMetric() throws Exception {
...
+	@Test
+	void testBasicPreDot11nAggregatedWifiScanEntry() {
+		final long obsoletionPeriodMs = 900000;
+
+		final String apA = "aaaaaaaaaaaa";
+		final String bssidA = "aa:aa:aa:aa:aa:aa";
+		final String apB = "bbbbbbbbbbbb";
+		final String bssidB = "bb:bb:bb:bb:bb:bb";
+		DeviceDataManager deviceDataManager = new DeviceDataManager();
+		deviceDataManager.setTopology(TestUtils.createTopology("test-zone", apA, apB));
```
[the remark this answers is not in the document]
RockyMandayam2: I'm actually not sure, I just saw it used in the beginning of other tests and copied it. I'll try it without this.

src/main/java/com/facebook/openwifirrm/modules/operationelement/OperationElement.java
elludraon: For now you can use "com.facebook.openwifirrm.ucentral.operationelement"

src/main/java/com/facebook/openwifirrm/modules/aggregators/Aggregator.java
elludraon: "com.facebook.openwifirrm.aggregators" for now

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -202,4 +212,110 @@ public static double calculateTPCMetrics(
...
+	 */
+	public static Map<String, Map<String, WifiScanEntry>> getAggregatedWifiScans(Modeler.DataModel dataModel,
+			long obsoletionPeriodMs,
+			Aggregator<Double> agg) {
+		if (obsoletionPeriodMs < 0) {
+			throw new IllegalArgumentException("obsoletionPeriodMs must be non-negative.");
+		}
+		Map<String, Map<String, WifiScanEntry>> aggregatedWifiScans = new HashMap<>();
+		for (Map.Entry<String, List<List<WifiScanEntry>>> apToScansMapEntry : dataModel.latestWifiScans.entrySet()) {
+			String ap = apToScansMapEntry.getKey();
```
elludraon, line 271: nit: we typically are calling this `serialNumber`, use same naming for consistency

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -202,4 +212,110 @@ public static double calculateTPCMetrics(
...
+					}).collect(Collectors.toUnmodifiableList());
+			// Split mostRecentToOldest into separate lists for each BSSID
+			// These lists will be in reverse chronological order also
+			Map<String, List<WifiScanEntry>> bssidToEntriesMap = new HashMap<>();
+			for (WifiScanEntry entry : mostRecentToOldestEntries) {
+				if (entry.bssid == null) {
+					continue;
+				}
+				bssidToEntriesMap.computeIfAbsent(entry.bssid, bssid -> new ArrayList<>());
+				bssidToEntriesMap.get(entry.bssid).add(entry);
```
elludraon, line 294: this is defeating the purpose of `computeIfAbsent` as you are doing a double lookup anyway with the following `get` -- replace with:

```java
bssidToEntriesMap
	.computeIfAbsent(entry.bssid, bssid -> new ArrayList<>())
	.add(entry);
```

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -202,4 +212,110 @@ public static double calculateTPCMetrics(
...
+				for (WifiScanEntry entry : entries) {
+					if (mostRecentEntry.unixTimeMs - entry.unixTimeMs > obsoletionPeriodMs) {
+						// discard obsolete entries
+						break;
+					}
+					// mostRecentEntry should always match the first entry (they are the exact same
+					// object)
+					if (matchesForAggregation(mostRecentEntry, entry)) {
+						aggregatedWifiScans.computeIfAbsent(ap, k -> new HashMap<>());
+						aggregatedWifiScans.get(ap).putIfAbsent(bssid, new WifiScanEntry(entry));
```
elludraon, line 312: - same here re: `computeIfAbsent` -> `get` doing double-lookup
- instead of `putIfAbsent` you should just use `computeIfAbsent` again, which will avoid unnecessarily calling the `new WifiScanEntry()` constructor

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -202,4 +212,110 @@ public static double calculateTPCMetrics(
...
+						agg.addValue((double) entry.signal);
+					}
+				}
+				aggregatedWifiScans.get(ap).get(bssid).signal = (int) Math.round(agg.getAggregate());
```
elludraon, line 316: are all map values here guaranteed to exist (seems not due to `obsoletionPeriodMs` check)? need to check for null?

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -202,4 +212,110 @@ public static double calculateTPCMetrics(
...
+			for (Map.Entry<String, List<WifiScanEntry>> bssidToWifiScanEntriesMapEntry : bssidToEntriesMap.entrySet()) {
+				String bssid = bssidToWifiScanEntriesMapEntry.getKey();
+				List<WifiScanEntry> entries = bssidToWifiScanEntriesMapEntry.getValue();
+				WifiScanEntry mostRecentEntry = entries.get(0);
+				agg.reset();
+				for (WifiScanEntry entry : entries) {
+					if (mostRecentEntry.unixTimeMs - entry.unixTimeMs > obsoletionPeriodMs) {
+						// discard obsolete entries
+						break;
+					}
```
elludraon, line 307: @pohanhf - how did you intend to define `obsoletionPeriodMs` ? I thought this was going to be measured from the current time when the algorithm is run, not based on the most recent per-BSSID scan time (as it is implemented here).

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -202,4 +212,110 @@ public static double calculateTPCMetrics(
...
+					// mostRecentEntry should always match the first entry (they are the exact same
+					// object)
+					if (matchesForAggregation(mostRecentEntry, entry)) {
```
elludraon, line 310: nit: can you explicitly write `if (mostRecentEntry == entry || matchesForAggregation(...))` instead? then comment is not needed either

src/test/java/com/facebook/openwifirrm/optimizers/TestUtils.java
```diff
@@ -26,6 +28,8 @@ public class TestUtils {
 	/** The Gson instance. */
 	private static final Gson gson = new Gson();
 
+	public static final Instant DEFAULT_WIFISCANENTRY_TIME = Instant.parse("2022-01-01T00:00:00Z");
```
elludraon, line 31: still needs comment

elludraon (changes requested):

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
```diff
@@ -83,7 +83,20 @@ public static class DataModel {
...
+		 * wifiscan request but instead just waits for periodic beacons from the other
+		 * APs. (Note that neither the responses to requests (in active mode) or the
+		 * periodic beacons are guaranteed to happen at any particular time (and it
+		 * depends on network traffic)).
+		 * <p>
+		 * The "result" of a wifiscan therefore can include multiple responses.
+		 * {@code latestWifiScans} maps from an AP (serial number) to a list of most
+		 * recent wifiscan "results" where each "result" itself is a list of responses
+		 * from other APs.
+		 */
```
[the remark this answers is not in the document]
RockyMandayam2: I wasn't really sure where to put it. I had a misunderstanding of this data structure earlier and thought I'd document what I know. Any suggestions for where to put it? Or should I just remove it?

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
RockyMandayam2: image
Po-Han wants the most recent per-BSSID scan time as I understand it based on this earlier comment chain from the closed PR (the one that was closed and split into two separate smaller PRs).

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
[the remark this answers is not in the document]
elludraon: top paragraph into `UCentralClient.wifiScan()` if you want to, bottom paragraph can stay here (reword it, e.g. remove "therefore")

src/test/java/com/facebook/openwifirrm/modules/aggregators/MeanAggregatorTest.java
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+package com.facebook.openwifirrm.modules.aggregators;
```
elludraon, line 9: move to `com.facebook.openwifirrm.aggregators`

src/test/java/com/facebook/openwifirrm/modules/operationelement/HTOperationElementTest.java
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+package com.facebook.openwifirrm.modules.operationelement;
```
elludraon, line 9: move to `com.facebook.openwifirrm.ucentral.operationelement`

src/test/java/com/facebook/openwifirrm/modules/operationelement/VHTOperationElementTest.java
```diff
@@ -0,0 +1,50 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+package com.facebook.openwifirrm.modules.operationelement;
```
elludraon, line 9: move to `com.facebook.openwifirrm.ucentral.operationelement`

elludraon (changes requested):
Almost done, you missed the test classes while refactoring

src/main/java/com/facebook/openwifirrm/modules/operationelement/HTOperationElement.java
RockyMandayam2: @elludraon do you want to wait for Po-Han to check this?

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
```diff
@@ -83,7 +83,14 @@ public static class DataModel {
 		// At minimum, we may want to aggregate recent wifi scan responses and
 		// keep a rolling average for stats.
 
-		/** List of latest wifi scan results per device. */
+		/**
+		 * The "result" of a wifiscan can include multiple responses.
+		 * {@code latestWifiScans} maps from an AP (serial number) to a list of most
```
elludraon, line 88: nit: don't put @ reference to self (the thing you're already commenting on)

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -202,4 +212,108 @@ public static double calculateTPCMetrics(
 			return Double.POSITIVE_INFINITY;
 		}
 	}
+
+	/**
+	 * Determines if two wifiscan entries should be aggregated without consideration
+	 * for any kind of obsoletion period. (Untested) should handle APs that support
```
elludraon, line 218: move "(Untested)" out of the javadoc and into an inline "TODO" comment, or open an internal task to track this

elludraon (approved):

paulvarkey (approved):

src/main/java/com/facebook/openwifirrm/ucentral/operationelement/HTOperationElement.java
```diff
@@ -0,0 +1,206 @@
...
+	 */
+	public HTOperationElement(String htOper) {
+		byte[] bytes = Base64.decodeBase64(htOper);
+		this.primaryChannel = bytes[0];
+		this.secondaryChannelOffset = (byte) (bytes[1] >>> 6);
+		this.staChannelWidth = ((bytes[1] & 0b00100000) >>> 5) == 1;
+		this.rifsMode = ((bytes[1] & 0b00010000) >>> 4) == 1;
+		this.htProtection = (byte) (bytes[2] >>> 6);
+		this.nongreenfieldHtStasPresent = ((bytes[2] & 0b00100000) >>> 5) == 1;
+		this.obssNonHtStasPresent = ((bytes[2] & 0b00001000) >>> 4) == 1;
```
halmehmood, line 127: should be >>> 3

src/main/java/com/facebook/openwifirrm/ucentral/operationelement/HTOperationElement.java
```diff
@@ -0,0 +1,206 @@
...
+				channelCenterFrequencySegment2, false, false, false);
+	}
+
+	/**
+	 * Constructs an {@code HTOperationElement} by decoding {@code htOper}.
+	 *
+	 * @param htOper a base64 encoded properly formatted HT operation element (see
+	 *               802.11)
+	 */
+	public HTOperationElement(String htOper) {
```
halmehmood, line 119: Are you sure that B0 is supposed to be the right-most bit? It looks so from the figures in the standard but just want to be sure.

halmehmood (commented):
Great work! Impressed that you did the extra effort of coding HT and VHT operation elements.

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
```diff
@@ -347,4 +346,22 @@ public static String generateServiceKey(RRMConfig.ServiceConfig serviceConfig) {
...
+
+	/**
+	 * Converts channel number to that channel's center frequency in MHz.
+	 *
+	 * @param channel channel number. See
+	 *                {@link ChannelOptimizer#AVAILABLE_CHANNELS_BAND} for channels
+	 *                in each band.
+	 * @return the center frequency of the given channel in MHz
+	 */
+	public static int channelToFrequencyMHz(int channel) {
```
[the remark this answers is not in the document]
RockyMandayam2: Currently it is only used for testing (specifically, to make more realistic wifiscan entries where both the `channel` field and the `frequency` fields are filled in).

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
```diff
@@ -41,78 +44,74 @@ public class UCentralUtils {
...
-		 * This field is specified as 24 bytes, but it is encoded in base64. It is
-		 * likely the case that the first byte (the Element ID, which should be 61 for
-		 * ht_oper) and the second byte (Length) are omitted in the wifi scan results,
-		 * resulting in 22 bytes, which translates to a 32 byte base64 encoded String.
-		 */
-		public String ht_oper;
+	/**
+	 * Extends {@link WifiScanEntryResult} to track the response time of the entry.
+	 */
+	public static class WifiScanEntry extends WifiScanEntryResult {
```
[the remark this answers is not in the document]
RockyMandayam2: The naming of these two classes has gone through a few rounds of re-evaluation. At some point, I had done `WifiScanEntry` and `ProcessedWifiScanEntry` but @elludraon did not like the name `ProcessedWifiScanEntry`. @elludraon what do you think of `WifiScanEntryTs` (or perhaps something like `TimestampedWifiScanEntry`)?

src/test/java/com/facebook/openwifirrm/modules/ModelerUtilsTest.java
```diff
@@ -77,4 +86,216 @@ void testValidMetric() throws Exception {
...
+		assertEquals(expectedAggregatedEntryAToB, aggregateMap.get(apB).get(bssidA));
+		// test moving the boundary by 1 ms and excluding the earliest entry
+		aggregateMap = ModelerUtils.getAggregatedWifiScans(dataModel, obsoletionPeriodMs - 1, new MeanAggregator());
+		expectedAggregatedEntryAToB.signal = -63; // average of -62 and -64
+		assertEquals(expectedAggregatedEntryAToB, aggregateMap.get(apB).get(bssidA));
+		// test an obsoletion period of 0 ms
+		aggregateMap = ModelerUtils.getAggregatedWifiScans(dataModel, 0, new MeanAggregator());
+		expectedAggregatedEntryAToB.signal = -64; // latest rssid
+		assertEquals(expectedAggregatedEntryAToB, aggregateMap.get(apB).get(bssidA));
+		// Test that the obsoletion period starts counting backwards from the time of
```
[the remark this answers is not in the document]
RockyMandayam2: What you suggested is what I had initially done, but @pohanhf suggested I do it starting from the most recent entry for each bssid.

src/test/java/com/facebook/openwifirrm/modules/ModelerUtilsTest.java
```diff
@@ -77,4 +86,216 @@ void testValidMetric() throws Exception {
...
+		// the most recent entry for each (ap, bssid) tuple.
+		WifiScanEntry entryCToB1 = TestUtils.createWifiScanEntryWithBssid(1, bssidC);
+		entryCToB1.signal = -70;
+		entryCToB1.unixTimeMs += 2 * obsoletionPeriodMs;
+		dataModel.latestWifiScans.get(apB).add(Arrays.asList(entryCToB1));
+		aggregateMap = ModelerUtils.getAggregatedWifiScans(dataModel, 0, new MeanAggregator());
```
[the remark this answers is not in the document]
RockyMandayam2: Yes. 0 indicates it should only use the most recent entry and nothing else. I am testing that it does this for each bssid for each AP (as opposed to something buggy, like only the most recent entry across all bssids for a given ap).

src/test/java/com/facebook/openwifirrm/modules/ModelerUtilsTest.java
```diff
@@ -77,4 +86,216 @@ void testValidMetric() throws Exception {
...
+		WifiScanEntry entryAToB3 = TestUtils.createWifiScanEntryWithWidth(bssidA, primaryChannel, htOper, null);
+		entryAToB3.signal = -70;
+		entryAToB3.unixTimeMs += 120000; // 1 min after previous entry
+		dataModel.latestWifiScans.get(apB).add(Arrays.asList(entryAToB3));
+		aggregateMap = ModelerUtils.getAggregatedWifiScans(dataModel, obsoletionPeriodMs, new MeanAggregator());
+		expectedAggregatedEntryAToB = new WifiScanEntry(entryAToB3);
+		expectedAggregatedEntryAToB.signal = -65; // average of -60 and -70 (would be -64 if the -62 entry was included)
+		assertEquals(expectedAggregatedEntryAToB, aggregateMap.get(apB).get(bssidA));
+
+		// test that entries with different HT operation elements are not aggregated
```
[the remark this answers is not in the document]
RockyMandayam2: Good catch, I will address this in my next push.

src/test/java/com/facebook/openwifirrm/modules/ModelerUtilsTest.java
```diff
@@ -77,4 +86,216 @@ void testValidMetric() throws Exception {
...
+		// even if their other fields match
```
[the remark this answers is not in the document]
RockyMandayam2: Will do in the next push

src/main/java/com/facebook/openwifirrm/ucentral/operationelement/OperationElement.java
```diff
@@ -0,0 +1,23 @@
...
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+package com.facebook.openwifirrm.ucentral.operationelement;
+
+/** Represents an Operation Element defined in 802.11. */
+public interface OperationElement {
```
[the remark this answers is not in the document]
RockyMandayam2: Fair, I will remove in the next push.

src/main/java/com/facebook/openwifirrm/ucentral/operationelement/HTOperationElement.java
RockyMandayam2: Good catch, I will fix it in my next push.

src/main/java/com/facebook/openwifirrm/ucentral/operationelement/HTOperationElement.java
RockyMandayam2: By B0 do you mean "bit 0" of the basic MCS set field?

If so, it doesn't actually matter whether it is right-most or left-most since right now I'm just copying it bit-by-bit (I copy the bytes exactly from the decoded String into `basicHtMcsSet`) and never using it. But if it did matter, I think usually bit 0 refers to the left-most and that is also what I see in Section 9.4.2.55.4.

src/main/java/com/facebook/openwifirrm/modules/operationelement/HTOperationElement.java
RockyMandayam2: Haleema checked it.

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
RockyMandayam2: I will change this.

src/test/java/com/facebook/openwifirrm/modules/ModelerUtilsTest.java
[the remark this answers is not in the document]
RockyMandayam2: I will add a comment explaining this

src/test/java/com/facebook/openwifirrm/modules/ModelerUtilsTest.java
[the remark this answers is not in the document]
RockyMandayam2: And also for vht_oper

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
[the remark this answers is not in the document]
elludraon: Whatever we decide, let's do it in a separate PR since that'll result in a lot of refactoring changes.

src/main/java/com/facebook/openwifirrm/ucentral/operationelement/HTOperationElement.java
```diff
@@ -137,14 +137,9 @@ public HTOperationElement(String htOper) {
...
 
-	@Override
-	public boolean matchesForAggregation(OperationElement otherOper) {
-		if (otherOper == null || getClass() == otherOper.getClass()) {
-			return false;
-		}
-		HTOperationElement other = (HTOperationElement) otherOper;
-		return primaryChannel == other.primaryChannel && secondaryChannelOffset == other.secondaryChannelOffset
-				&& staChannelWidth == other.staChannelWidth
+	public boolean matchesForAggregation(HTOperationElement other) {
```
elludraon, line 140: add javadoc comment if you've removed `@Override`

src/main/java/com/facebook/openwifirrm/ucentral/operationelement/VHTOperationElement.java
```diff
@@ -90,14 +90,9 @@ public VHTOperationElement(boolean channelWidthIndicator, byte channel1, byte ch
 		this.vhtMcsForNss = vhtMcsForNss;
 	}
 
-	@Override
-	public boolean matchesForAggregation(OperationElement otherOper) {
+	public boolean matchesForAggregation(VHTOperationElement other) {
```
elludraon, line 93: add javadoc comment if you've removed `@Override`

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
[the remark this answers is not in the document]
RockyMandayam2: Ok. Since Po-Han had thoughts too, I will wait till he's back from leave before I start a new conversation about this :)

RockyMandayam2:
> Great work! Impressed that you did the extra effort of coding HT and VHT operation elements.

Thank you!

src/test/java/com/facebook/openwifirrm/modules/ModelerUtilsTest.java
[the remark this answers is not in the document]
RockyMandayam2: Haleema said we can defer to Po-Han for now

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
halmehmood: Yes, this is what I think makes most sense

src/main/java/com/facebook/openwifirrm/ucentral/operationelement/HTOperationElement.java
RockyMandayam2: Haleema and I discussed. She made a good catch. The actual operation elements come in "big endian" in terms of byte order but "little endian" in terms of bit order within a byte (alternatively, one can interpret it as all "big endian" but in that case the illustrations in the 802.11 are drawn "backwards" graphically). I will fix this in the next push.

halmehmood (approved):

PR #30, opened 2022-08-15
Handle empty topology file

RockyMandayam2:
Handle blank input files (blank meaning containing no characters or only whitespace), which was not handled before. Only "empty JSONs" (i.e., `{}`) were handled before.

Signed-off-by: Rocky Mandayam <rockymandayam@fb.com>

### DIFF
--- a/src/main/java/com/facebook/openwifirrm/DeviceDataManager.java
+++ b/src/main/java/com/facebook/openwifirrm/DeviceDataManager.java
@@ -114,7 +114,6 @@ public class DeviceDataManager {
 			} else {
 				topo = gson.fromJson(contents, DeviceTopology.class);
 			}
-
 		}
 		validateTopology(topo);
 		return topo;
--- a/src/main/java/com/facebook/openwifirrm/DeviceDataManager.java
+++ b/src/main/java/com/facebook/openwifirrm/DeviceDataManager.java
@@ -106,7 +106,15 @@ public class DeviceDataManager {
 			// Read file
 			logger.info("Reading topology file '{}'", topologyFile.getPath());
 			String contents = Utils.readFile(topologyFile);
-			topo = gson.fromJson(contents, DeviceTopology.class);
+			if (contents == null || contents.isBlank()) {
+				// Blank file, write defaults to disk
+				logger.info("Topology file '{}' is blank, writing to it...", topologyFile.getPath());
+				topo = new DeviceTopology();
+				Utils.writeJsonFile(topologyFile, topo);
+			} else {
+				topo = gson.fromJson(contents, DeviceTopology.class);
+			}
+
 		}
 		validateTopology(topo);
 		return topo;
--- a/src/main/java/com/facebook/openwifirrm/DeviceDataManager.java
+++ b/src/main/java/com/facebook/openwifirrm/DeviceDataManager.java
@@ -129,7 +129,7 @@ public class DeviceDataManager {
 		}
 		if (cfg == null) {
 			// Missing/empty file, write defaults to disk
-			logger.info("Creating default config file '{}'", deviceConfigFile.getPath());
+			logger.info("Creating default device config file '{}'", deviceConfigFile.getPath());
 			cfg = new DeviceLayeredConfig();
 			Utils.writeJsonFile(deviceConfigFile, cfg);
 		} else {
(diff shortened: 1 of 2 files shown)


src/main/java/com/facebook/openwifirrm/DeviceDataManager.java
```diff
@@ -106,7 +106,13 @@ private DeviceTopology readTopology(File topologyFile) throws IOException {
 			// Read file
 			logger.info("Reading topology file '{}'", topologyFile.getPath());
 			String contents = Utils.readFile(topologyFile);
-			topo = gson.fromJson(contents, DeviceTopology.class);
+			if (contents == null || contents.isBlank()) {
```
elludraon, line 109: - `Utils.readFile()` will never return null, don't check for it
- Use `.isEmpty()` instead of `.isBlank()`

However this whole block seems too verbose, you can write it much more succinctly (without the branching and duplicate logic):
```java
		DeviceTopology topo = null;
		if (topologyFile.isFile()) {
			// Read file
			logger.info("Reading topology file '{}'", topologyFile.getPath());
			String contents = Utils.readFile(topologyFile);
			topo = gson.fromJson(contents, DeviceTopology.class);
		}
		if (topo == null) {
			// Missing/empty file, write defaults to disk
			logger.info(
				"Creating default topology file '{}'", topologyFile.getPath()
			);
			topo = new DeviceTopology();
			Utils.writeJsonFile(topologyFile, topo);
		}
```
Note this is possible since `gson.fromJson()` properly handles empty strings and returns null, so you don't need to check it yourself.

Please do the same in the other methods too.

src/main/java/com/facebook/openwifirrm/Launcher.java
```diff
@@ -82,21 +82,24 @@ private RRMConfig readRRMConfig(File configFile) throws IOException {
...
-			Utils.jsonMerge(fullConfig, userConfig);
-			config = gson.fromJson(fullConfig.toString(), RRMConfig.class);
-
-			// Compare merged config with contents as read from disk
-			// If any differences (ex. added fields), overwrite config file
-			if (!fullConfig.toString().equals(userConfig.toString())) {
-				logger.info("Rewriting config file with new changes...");
-				try (Writer writer = new FileWriter(configFile)) {
-					gson.toJson(config, writer);
+			if (contents == null || contents.isBlank()) {
```
elludraon, line 85: same here, flip the outer conditions but you'll need to do the `contents.isEmpty()` check yourself (`new JSONObject` doesn't handle it)

src/main/java/com/facebook/openwifirrm/Launcher.java
```diff
@@ -82,21 +82,24 @@ private RRMConfig readRRMConfig(File configFile) throws IOException {
 			// Read file
 			logger.info("Reading config file '{}'", configFile.getPath());
 			String contents = Utils.readFile(configFile);
-			JSONObject userConfig = new JSONObject(contents);
-
```
elludraon, line 86: please leave in these blank lines for spacing (it was intended to create "paragraphs" of comments+code)

elludraon (changes requested):

src/main/java/com/facebook/openwifirrm/DeviceDataManager.java
RockyMandayam2: Will do. This will result in the log not showing whether the file is missing vs blank, but I suppose this isn't a big deal.

src/main/java/com/facebook/openwifirrm/DeviceDataManager.java
```diff
@@ -121,23 +120,19 @@ private DeviceTopology readTopology(File topologyFile) throws IOException {
...
-				"Reading device config file '{}'", deviceConfigFile.getPath()
-			);
+			logger.info("Reading device config file '{}'", deviceConfigFile.getPath());
 			String contents = Utils.readFile(deviceConfigFile);
 			cfg = gson.fromJson(contents, DeviceLayeredConfig.class);
-
+		}
+		if (cfg == null) {
+			// Missing/empty file, write defaults to disk
+			logger.info("Creating default config file '{}'", deviceConfigFile.getPath());
```
elludraon, line 132: "config file" => "device config file" (otherwise it's ambiguous w/ `RRMConfig`)

elludraon (approved):
Much better! One last comment, then go ahead and push.

PR #31, opened 2022-08-16
Configure javadoc, fix errors, deploy to gh-pages

elludraon:
- Add Javadoc config section to `pom.xml`
- Fix Javadoc errors in sources
- Add minimal `overview.html` to be placed on generated index page
- Add GitHub workflow to build Javadoc on push and deploy to gh-pages branch

### DIFF
--- a/src/main/java/com/facebook/openwifirrm/RRMAlgorithm.java
+++ b/src/main/java/com/facebook/openwifirrm/RRMAlgorithm.java
@@ -62,13 +62,13 @@ public class RRMAlgorithm {
 
 		/**
 		 * Computed channel assignments.
-		 * @see {@link ChannelOptimizer#computeChannelMap()}
+		 * @see ChannelOptimizer#computeChannelMap()
 		 */
 		public Map<String, Map<String, Integer>> channelMap;
 
 		/**
 		 * Computed tx power assignments.
-		 * @see {@link TPC#computeTxPowerMap()}
+		 * @see TPC#computeTxPowerMap()
 		 */
 		public Map<String, Map<String, Integer>> txPowerMap;
 	}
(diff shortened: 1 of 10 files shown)


src/main/javadoc/overview.html
```diff
@@ -0,0 +1,9 @@
+<!DOCTYPE HTML>
+<html lang="en">
+  <body>
+    <p>
+      OpenWiFi uCentral-based radio resource management service. Sources are available
```
RockyMandayam2, line 5: minor: maybe add a " (RRM)" after "radio resource management"

src/main/java/com/facebook/openwifirrm/RRMAlgorithm.java
```diff
@@ -98,7 +98,7 @@ public RRMAlgorithm(String name, Map<String, String> args) {
 	 *
 	 * @param name the algorithm name, which must exist in {@link AlgorithmType}
 	 * @param argsRaw the arguments as a comma-separated list of key=value
-	 *                pairs, e.g. <tt> key1=val1,key2=val2,key3=val3</tt>
+	 *                pairs, e.g. <code> key1=val1,key2=val2,key3=val3</code>
```
RockyMandayam2, line 101: Should I be worried about when to use `<code>` vs `{@code ...}`? I've always used `{@code ...}` since that's the only one I knew of to indicate that something is "source code"

RockyMandayam2 (approved):

src/main/java/com/facebook/openwifirrm/RRMAlgorithm.java
elludraon: Oh you're right, I should have used `{@code ...}`, will update.

PR #32, opened 2022-08-16
[WIFI-9985] Add CORS support

elludraon:
Add proper CORS support to `ApiServer`.
- Add global HTTP `OPTIONS` handler to reflect requested options.
- Add "Access-Control-Allow-Origin" response header (and "Vary") whenever "Origin" request header is present, *if* CORS checks pass based on `APISERVERPARAMS_CORSDOMAINLIST` RRM config:
    - If set, origin must match a domain in this comma-separated list.
    - If empty, always reflect the "Origin" header (i.e. allow all origins).

### DIFF
--- a/src/main/java/com/facebook/openwifirrm/modules/ApiServer.java
+++ b/src/main/java/com/facebook/openwifirrm/modules/ApiServer.java
@@ -428,7 +428,7 @@ public class ApiServer implements Runnable {
 			response.header("Access-Control-Allow-Methods", acrm);
 		}
 		response.header("Access-Control-Allow-Credentials", "true");
-		response.header("Access-Control-Max-Age", "86400");
+		response.header("Access-Control-Max-Age", "86400" /* 1 day */);
 		return "OK";
 	}
 
--- a/src/main/java/com/facebook/openwifirrm/modules/ApiServer.java
+++ b/src/main/java/com/facebook/openwifirrm/modules/ApiServer.java
@@ -9,6 +9,8 @@
 package com.facebook.openwifirrm.modules;
 
 import java.net.InetAddress;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.net.UnknownHostException;
 import java.time.Instant;
 import java.util.Arrays;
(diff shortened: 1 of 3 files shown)


src/main/java/com/facebook/openwifirrm/modules/ApiServer.java
```diff
@@ -338,6 +362,26 @@ private void beforeFilter(Request request, Response response) {
...
+		String origin = request.headers("Origin");
+		if (origin != null) {
+			boolean allowOrigin = false;
+			if (corsDomains == null) {
+				allowOrigin = true;
+			} else {
+				try {
+					String originHost = new URI(origin).getHost();
+					allowOrigin = corsDomains.contains(originHost);
+				} catch (URISyntaxException e) { /* invalid origin URI */ }
```
RockyMandayam2, line 375: Should there be some kind of log message here?

src/main/java/com/facebook/openwifirrm/modules/ApiServer.java
```diff
@@ -372,6 +416,22 @@ private void afterFilter(Request request, Response response) {
...
+		String acrh = request.headers("Access-Control-Request-Headers");
+		if (acrh != null) {
+			response.header("Access-Control-Allow-Headers", acrh);
+		}
+		String acrm = request.headers("Access-Control-Request-Method");
+		if (acrm != null) {
+			response.header("Access-Control-Allow-Methods", acrm);
+		}
+		response.header("Access-Control-Allow-Credentials", "true");
+		response.header("Access-Control-Max-Age", "86400");
```
RockyMandayam2, line 431: Maybe make 86400 a parameter

RockyMandayam2, line 431: Maybe add a comment saying it is 86400 seconds -> 24 hours.

RockyMandayam2 (approved):

src/main/java/com/facebook/openwifirrm/modules/ApiServer.java
elludraon: This gets captured in the "CORS checks failed" log below

src/main/java/com/facebook/openwifirrm/modules/ApiServer.java
elludraon: I decided not to since this isn't important to change, it's just the cache time

elludraon: OK

PR #33, opened 2022-08-17
Update package version to 2.7.0, append git info during docker builds

elludraon:
Update package version from 1.0.0 to 2.7.0, both in `pom.xml` and `ApiServer.java`.

Define a new `appendVersionString` property in Maven which gets added to the generated (filtered) `version.txt` resource (used to source the version string at runtime). In the `Dockerfile`, set this property to a combination of Git hash and build number when available (via new script `scripts/get_build_version.sh`).

Example output:
```
$ docker build -f Dockerfile .
Successfully built 0c3cabe843d1

$ docker run --rm -it 0c3cabe843d1 /bin/bash

# java -jar /usr/local/bin/openwifi-rrm.jar -V
2.7.0-ebed9b5(32)
```

### DIFF
--- a/src/main/java/com/facebook/openwifirrm/modules/ApiServer.java
+++ b/src/main/java/com/facebook/openwifirrm/modules/ApiServer.java
@@ -89,7 +89,7 @@ import spark.Spark;
 @OpenAPIDefinition(
 	info = @Info(
 		title = "OpenWiFi 2.0 RRM OpenAPI",
-		version = "1.0.0",  // TODO
+		version = "2.7.0",  // NOTE: needs manual update!
 		description = "This document describes the API for the Radio Resource Management service."
 	),
 	tags = {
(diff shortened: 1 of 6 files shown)


src/main/java/com/facebook/openwifirrm/modules/ApiServer.java
```diff
@@ -89,7 +89,7 @@
 @OpenAPIDefinition(
 	info = @Info(
 		title = "OpenWiFi 2.0 RRM OpenAPI",
-		version = "1.0.0",  // TODO
+		version = "2.7.0",  // NOTE: needs manual update!
```
RockyMandayam2, line 92: Minor: Don't know much about the convention, but maybe `XXX` is even better than `NOTE`?

RockyMandayam2 (approved):

src/main/java/com/facebook/openwifirrm/modules/ApiServer.java
elludraon: Going to leave this since it seems nobody on the internet can agree what `XXX` is supposed to mean - https://stackoverflow.com/q/1452934

PR #42, opened 2022-08-19
Add missing device connection status checks in DataCollector

elludraon:
In `DataCollector`, check if a device is online/connected before trying to send a wifi scan or capabilities request. (This is already checked in `ConfigManager`.)

### DIFF
--- a/src/main/java/com/facebook/openwifirrm/modules/DataCollector.java
+++ b/src/main/java/com/facebook/openwifirrm/modules/DataCollector.java
@@ -260,6 +260,15 @@ public class DataCollector implements Runnable {
 			Math.max(params.capabilitiesIntervalSec, 1) * 1_000_000_000L;
 
 		for (DeviceWithStatus device : devices) {
+			// Check if online/connected
+			if (!device.connected) {
+				logger.info(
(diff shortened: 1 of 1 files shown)


src/main/java/com/facebook/openwifirrm/modules/DataCollector.java
```diff
@@ -260,6 +260,15 @@ private void fetchDeviceCapabilities(List<DeviceWithStatus> devices) {
 			Math.max(params.capabilitiesIntervalSec, 1) * 1_000_000_000L;
 
 		for (DeviceWithStatus device : devices) {
+			// Check if online/connected
```
RockyMandayam2, line 263: I think the code `!device.connected` is self-explanatory, so you don't need the comment. Up to you. Same comment for the other instances of `!device.connected`

RockyMandayam2 (approved):

src/main/java/com/facebook/openwifirrm/modules/DataCollector.java
elludraon: I like to keep comments above nearly every "paragraph" of code so it's quicker to skim files without actually reading any code.

PR #47, opened 2022-08-25
[WIFI-10593] Allow different values per AP for `RandomChannelInitializer` and `RandomTxPowerInitializer`

jwoos:
Previously both `RandomChannelInitializer` and `RandomTxPowerInitializer` would initialize a single random value to use across APs. Introduce another constructor to allow setting a different random value PER AP.

In addition:
- Make  `RandomTxPowerInitializer` follow `RandomChannelInitializer`'s lead in terms of how random values are generated. Instead of holding an attribute for the generated value, it'll be generated on the fly since the value isn't used outside of computing the map.
- Add another constructor that also allows passing in a `Random` instance to allow seeding, making testing possible by having the number generation deterministic. I didn't expose a method for setting the seed as that would introduce weird cases where numbers could have been generated but then the seed was changed.

[WIFI-10593](https://telecominfraproject.atlassian.net/browse/WIFI-10593)

### DIFF
--- a/src/main/java/com/facebook/openwifirrm/optimizers/channel/RandomChannelInitializer.java
+++ b/src/main/java/com/facebook/openwifirrm/optimizers/channel/RandomChannelInitializer.java
@@ -38,28 +38,29 @@ public class RandomChannelInitializer extends ChannelOptimizer {
 	public static final String ALGORITHM_ID = "random";
 
 	/** The PRNG instance. */
-	private Random rng = new Random();
+	private final Random rng;
 
 	/** Whether to set a different value per AP or use a single value for all APs */
-	private boolean setDifferentChannelsPerAp = false;
+	private final boolean setDifferentChannelsPerAp;
 
 	/** Constructor. */
 	public RandomChannelInitializer(
 		DataModel model, String zone, DeviceDataManager deviceDataManager
 	) {
-		super(model, zone, deviceDataManager);
+		this(model, zone, deviceDataManager, false, new Random());
 	}
 
 	/** Constructor (allows setting different channel per AP) */
 	public RandomChannelInitializer(
 		DataModel model, String zone, DeviceDataManager deviceDataManager, boolean setDifferentChannelsPerAp
 	) {
-		this(model, zone, deviceDataManager);
-		this.setDifferentChannelsPerAp = setDifferentChannelsPerAp;
+		this(model, zone, deviceDataManager, setDifferentChannelsPerAp, new Random());
 	}
 
-	/** Constructor (allows setting different channel per AP and passing
-	 * in a custom Random class to allow seeding) */
+	/**
+	 * Constructor (allows setting different channel per AP and passing
+	 * in a custom Random class to allow seeding)
+	 */
 	public RandomChannelInitializer(
 		DataModel model,
 		String zone,
@@ -67,7 +68,8 @@ public class RandomChannelInitializer extends ChannelOptimizer {
 		boolean setDifferentChannelsPerAp,
 		Random rng
 	) {
-		this(model, zone, deviceDataManager, setDifferentChannelsPerAp);
+		super(model, zone, deviceDataManager);
+		this.setDifferentChannelsPerAp = setDifferentChannelsPerAp;
 		this.rng = rng;
 	}
 
@@ -119,7 +121,9 @@ public class RandomChannelInitializer extends ChannelOptimizer {
 				);
 			}
 
-			// Randomly assign all the devices to the same channel
+			// Randomly assign all the devices to the same channel if
+			// setDifferentChannelsPerAp is false otherwise, assigns
+			// each device to a random channel
 			int defaultChannelIndex = rng.nextInt(availableChannelsList.size());
 
 			for (String serialNumber : entry.getValue()) {
--- a/src/main/java/com/facebook/openwifirrm/optimizers/channel/RandomChannelInitializer.java
+++ b/src/main/java/com/facebook/openwifirrm/optimizers/channel/RandomChannelInitializer.java
@@ -43,20 +43,6 @@ public class RandomChannelInitializer extends ChannelOptimizer {
 	/** Whether to set a different value per AP or use a single value for all APs */
 	private final boolean setDifferentChannelsPerAp;
 
-	/** Constructor. */
-	public RandomChannelInitializer(
-		DataModel model, String zone, DeviceDataManager deviceDataManager
-	) {
-		this(model, zone, deviceDataManager, false, new Random());
-	}
-
-	/** Constructor (allows setting different channel per AP) */
-	public RandomChannelInitializer(
-		DataModel model, String zone, DeviceDataManager deviceDataManager, boolean setDifferentChannelsPerAp
-	) {
-		this(model, zone, deviceDataManager, setDifferentChannelsPerAp, new Random());
-	}
-
 	/**
 	 * Constructor (allows setting different channel per AP and passing
 	 * in a custom Random class to allow seeding)
@@ -71,6 +57,20 @@ public class RandomChannelInitializer extends ChannelOptimizer {
 		super(model, zone, deviceDataManager);
 		this.setDifferentChannelsPerAp = setDifferentChannelsPerAp;
 		this.rng = rng;
+	}
+
+	/** Constructor (allows setting different channel per AP) */
+	public RandomChannelInitializer(
+		DataModel model, String zone, DeviceDataManager deviceDataManager, boolean setDifferentChannelsPerAp
+	) {
+		this(model, zone, deviceDataManager, setDifferentChannelsPerAp, new Random());
+	}
+
+	/** Constructor. */
+	public RandomChannelInitializer(
+		DataModel model, String zone, DeviceDataManager deviceDataManager
+	) {
+		this(model, zone, deviceDataManager, false);
 	}
 
 	@Override
--- a/src/main/java/com/facebook/openwifirrm/optimizers/channel/RandomChannelInitializer.java
+++ b/src/main/java/com/facebook/openwifirrm/optimizers/channel/RandomChannelInitializer.java
@@ -28,7 +28,8 @@ import com.facebook.openwifirrm.ucentral.models.State;
 /**
  * Random channel initializer.
  * <p>
- * Randomly assign APs to the same channel.
+ * Randomly assign APs to the same channel unless otherwise specified.
+ * If specified, all APs will be assigned a random channel.
  */
 public class RandomChannelInitializer extends ChannelOptimizer {
 	private static final Logger logger = LoggerFactory.getLogger(RandomChannelInitializer.class);
@@ -37,13 +38,37 @@ public class RandomChannelInitializer extends ChannelOptimizer {
 	public static final String ALGORITHM_ID = "random";
 
 	/** The PRNG instance. */
-	private final Random rng = new Random();
+	private Random rng = new Random();
+
+	/** Whether to set a different value per AP or use a single value for all APs */
+	private boolean setDifferentChannelsPerAp = false;
 
 	/** Constructor. */
 	public RandomChannelInitializer(
 		DataModel model, String zone, DeviceDataManager deviceDataManager
 	) {
 		super(model, zone, deviceDataManager);
+	}
+
+	/** Constructor (allows setting different channel per AP) */
+	public RandomChannelInitializer(
+		DataModel model, String zone, DeviceDataManager deviceDataManager, boolean setDifferentChannelsPerAp
+	) {
+		this(model, zone, deviceDataManager);
+		this.setDifferentChannelsPerAp = setDifferentChannelsPerAp;
+	}
+
+	/** Constructor (allows setting different channel per AP and passing
+	 * in a custom Random class to allow seeding) */
+	public RandomChannelInitializer(
+		DataModel model,
+		String zone,
+		DeviceDataManager deviceDataManager,
+		boolean setDifferentChannelsPerAp,
+		Random rng
+	) {
+		this(model, zone, deviceDataManager, setDifferentChannelsPerAp);
+		this.rng = rng;
 	}
 
 	@Override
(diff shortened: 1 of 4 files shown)


src/test/java/com/facebook/openwifirrm/optimizers/tpc/RandomTxPowerInitializerTest.java
```diff
@@ -98,4 +117,24 @@ void testRandomTxPower() throws Exception {
...
+		);
+		Map<String, Map<String, Integer>> txPowerMap =
+			optimizer.computeTxPowerMap();
+		Set<Integer> txPowerSet = new TreeSet<>();
+		for (String serialNumber : txPowerMap.keySet()) {
+			txPowerSet.addAll(txPowerMap.get(serialNumber).values());
+		}
+
+		// different values per AP
+		assertNotEquals(1, txPowerSet.size());
```
RockyMandayam2, line 135: @elludraon do you think this test should be more specific than just testing that the set has more than one element?

src/main/java/com/facebook/openwifirrm/optimizers/channel/RandomChannelInitializer.java
```diff
@@ -46,6 +50,27 @@ public RandomChannelInitializer(
 		super(model, zone, deviceDataManager);
 	}
 
+	/** Constructor (allows setting different channel per AP) */
+	public RandomChannelInitializer(
+		DataModel model, String zone, DeviceDataManager deviceDataManager, boolean setDifferentChannelsPerAp
+	) {
+		this(model, zone, deviceDataManager);
```
elludraon, line 57: I am pretty sure you wrote this idiom backwards (based on what I've seen). It should go this way, which looks cleaner:
```java
public class X {
	private final int a, b;
	public X() {
		this(1);
	}
	public X(int a) {
		this(a, 2);
	}
	public X(int a, int b) {
		this.a = a;
		this.b = b;
	}
}
```

Also please fix this in the other class as well.

src/main/java/com/facebook/openwifirrm/optimizers/channel/RandomChannelInitializer.java
```diff
@@ -37,7 +38,10 @@ public class RandomChannelInitializer extends ChannelOptimizer {
 	public static final String ALGORITHM_ID = "random";
 
 	/** The PRNG instance. */
-	private final Random rng = new Random();
+	private Random rng = new Random();
```
elludraon, line 41: I would keep the `final` and instantiate it in the constructors, see comment below.

src/main/java/com/facebook/openwifirrm/optimizers/channel/RandomChannelInitializer.java
```diff
@@ -95,10 +120,14 @@ public Map<String, Map<String, Integer>> computeChannelMap() {
 			}
 
 			// Randomly assign all the devices to the same channel
```
elludraon, line 122: update this comment

src/main/java/com/facebook/openwifirrm/optimizers/channel/RandomChannelInitializer.java
```diff
@@ -46,6 +50,27 @@ public RandomChannelInitializer(
...
+	/** Constructor (allows setting different channel per AP) */
+	public RandomChannelInitializer(
+		DataModel model, String zone, DeviceDataManager deviceDataManager, boolean setDifferentChannelsPerAp
+	) {
+		this(model, zone, deviceDataManager);
+		this.setDifferentChannelsPerAp = setDifferentChannelsPerAp;
+	}
+
+	/** Constructor (allows setting different channel per AP and passing
+	 * in a custom Random class to allow seeding) */
```
elludraon, line 62: (since we've disabled comment autoformatting...)
If a Javadoc comment does not fit in one line, keep the `/**` and `*/` on their own separate lines.
```java
/**
 * Long comment
 */
```

src/main/java/com/facebook/openwifirrm/optimizers/tpc/RandomTxPowerInitializer.java
```diff
@@ -30,50 +31,60 @@ public class RandomTxPowerInitializer extends TPC {
...
-	public RandomTxPowerInitializer(
-		DataModel model, String zone, DeviceDataManager deviceDataManager
-	) {
-		this(model, zone, deviceDataManager, getRandomTxPower());
-	}
+	/** Whether to set a different value per AP or use a single value for all APs */
+	private boolean setDifferentTxPowerPerAp = false;
 
-	/** Constructor. */
+	/** Constructor (uses random tx power per AP and allows passing in a custom Random class to allow seeding). */
```
elludraon, line 40: 80 chars per line

src/test/java/com/facebook/openwifirrm/optimizers/channel/RandomChannelInitializerTest.java
```diff
@@ -63,4 +65,44 @@ void test1() throws Exception {
...
+
+		DeviceDataManager deviceDataManager = new DeviceDataManager();
+		deviceDataManager.setTopology(
+			TestUtils.createTopology(TEST_ZONE, deviceA, deviceB)
+		);
+
+		// A and B will be assigned to the same channel
+		DataModel dataModel = new DataModel();
+		dataModel.latestState.put(
+			deviceA, TestUtils.createState(6, channelWidth, "ddd")
```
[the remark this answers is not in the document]
elludraon: Yeah, and also add it to the string constants at the top. BSSID can also match the device serial number, e.g. `aaaaaaaaaaaa` serial -> `aa:aa:aa:aa:aa:aa` BSSID.

src/test/java/com/facebook/openwifirrm/optimizers/tpc/RandomTxPowerInitializerTest.java
```diff
@@ -63,16 +66,32 @@ private static DataModel createModel() {
...
 	}
 
+	/**
+	 * Return the first value provided by Random, for each key seed
+	 */
+	private static Map<Integer, Integer> mapRngSeedToFirstValue() {
+		Map<Integer, Integer> map = new HashMap<>();
+		map.put(0, 2);
+		map.put(10, 9);
+		map.put(100, 10);
```
elludraon, line 76: nit: can you just put this into the test method? it is more confusing to me that this function exists at all

src/test/java/com/facebook/openwifirrm/optimizers/tpc/RandomTxPowerInitializerTest.java
```diff
@@ -63,16 +66,32 @@ private static DataModel createModel() {
...
+	}
+
 	@Test
 	@Order(1)
-	void testProvidedTxPower() throws Exception {
-		for (int txPower : Arrays.asList(-1, 16, 27)) {
+	void testSeededTxPower() throws Exception {
+		Map<Integer, Integer> seedMap = mapRngSeedToFirstValue();
+
+		for (Map.Entry<Integer, Integer> entry : seedMap.entrySet()) {
```
elludraon, line 86: I don't think you need to try to make this test look like what it did before... it's fine to not loop or hardcode any seed -> value pairs, and instead just assert that all devices/radios got the same allocated txPower (whatever that might be).

src/test/java/com/facebook/openwifirrm/optimizers/tpc/RandomTxPowerInitializerTest.java
elludraon: This should be `assertTrue(txPowerSet.size() > 1)` because you don't want to pass on size=0.

elludraon (changes requested):

src/main/java/com/facebook/openwifirrm/optimizers/tpc/RandomTxPowerInitializer.java
```diff
@@ -30,50 +31,60 @@ public class RandomTxPowerInitializer extends TPC {
...
+		return rng.nextInt(TPC.MAX_TX_POWER + 1 - TPC.MIN_TX_POWER) + TPC.MIN_TX_POWER;
 	}
 
 	@Override
 	public Map<String, Map<String, Integer>> computeTxPowerMap() {
+		int defaultTxPower = this.getRandomTxPower();
+		logger.info("Default power: {}", defaultTxPower);
 		Map<String, Map<String, Integer>> txPowerMap = new TreeMap<>();
 		for (String serialNumber : model.latestState.keySet()) {
+			int txPower = setDifferentTxPowerPerAp ? getRandomTxPower() : defaultTxPower;
```
wuweicaifb, line 78: @pohanhf  I wonder if we need to set different tx power per channel/band?

src/main/java/com/facebook/openwifirrm/optimizers/tpc/RandomTxPowerInitializer.java
```diff
@@ -11,6 +11,7 @@
 import java.util.Map;
 import java.util.Random;
 import java.util.TreeMap;
+import java.util.LinkedList;
```
wuweicaifb, line 14: It isn't used anywhere.

@elludraon Is there a way to auto delete unused imports?

src/test/java/com/facebook/openwifirrm/optimizers/tpc/RandomTxPowerInitializerTest.java
```diff
@@ -63,16 +66,32 @@ private static DataModel createModel() {
 		return dataModel;
 	}
 
+	/**
+	 * Return the first value provided by Random, for each key seed
+	 */
+	private static Map<Integer, Integer> mapRngSeedToFirstValue() {
+		Map<Integer, Integer> map = new HashMap<>();
+		map.put(0, 2);
```
wuweicaifb, line 74: I would prefer TreeMap in tests if you intend to keep this map, so that the orders are reserved.

src/test/java/com/facebook/openwifirrm/optimizers/channel/RandomChannelInitializerTest.java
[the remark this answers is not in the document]
jwoos: got it - yeah I just sort of copied theses from the existing cases. Will update all of them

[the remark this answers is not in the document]
RockyMandayam2: If there are a lot like this, consider updating all the existing ones in a separate PR (e.g., if 10 files have this problem, we probably don't want to clutter this PR).

src/main/java/com/facebook/openwifirrm/optimizers/tpc/RandomTxPowerInitializer.java
```diff
@@ -30,50 +31,60 @@ public class RandomTxPowerInitializer extends TPC {
...
 	}
 
 	@Override
 	public Map<String, Map<String, Integer>> computeTxPowerMap() {
+		int defaultTxPower = this.getRandomTxPower();
+		logger.info("Default power: {}", defaultTxPower);
 		Map<String, Map<String, Integer>> txPowerMap = new TreeMap<>();
 		for (String serialNumber : model.latestState.keySet()) {
+			int txPower = setDifferentTxPowerPerAp ? getRandomTxPower() : defaultTxPower;
 			Map<String, Integer> radioMap = new TreeMap<>();
```
RockyMandayam2, line 79: Maybe just personal preference, I think I prefer
```
txPowerMap
    .computeIfAbsent(serialNumber, k -> new TreeMap<>()).
    put(band, txPower);
```

src/test/java/com/facebook/openwifirrm/optimizers/tpc/RandomTxPowerInitializerTest.java
jwoos: good to know! first time writing java :p

src/main/java/com/facebook/openwifirrm/optimizers/tpc/RandomTxPowerInitializer.java
jwoos: hmm i like the current implementation - it's clearer that the map is created for each loop since the keys are from a map and have to be unique

src/main/java/com/facebook/openwifirrm/optimizers/channel/RandomChannelInitializer.java
```diff
@@ -37,13 +38,39 @@ public class RandomChannelInitializer extends ChannelOptimizer {
...
+	private final Random rng;
+
+	/** Whether to set a different value per AP or use a single value for all APs */
+	private final boolean setDifferentChannelsPerAp;
 
 	/** Constructor. */
 	public RandomChannelInitializer(
 		DataModel model, String zone, DeviceDataManager deviceDataManager
+	) {
+		this(model, zone, deviceDataManager, false, new Random());
```
elludraon, line 50: nit: drop the last argument here (`new Random()`) as you can just keep chaining down, which is cleaner code IMO

wuweicaifb (approved):

src/main/java/com/facebook/openwifirrm/optimizers/tpc/RandomTxPowerInitializer.java
```diff
@@ -30,50 +30,63 @@ public class RandomTxPowerInitializer extends TPC {
...
+		DataModel model, String zone, DeviceDataManager deviceDataManager, boolean setDifferentTxPowerPerAp
+	) {
+		this(model, zone, deviceDataManager, setDifferentTxPowerPerAp, new Random());
+	}
+
+	/** Constructor (uses random tx power). */
+	public RandomTxPowerInitializer(
+		DataModel model, String zone, DeviceDataManager deviceDataManager
+	) {
+		this(model, zone, deviceDataManager, false, new Random());
```
elludraon, line 66: nit: put all these constructors in order from least -> most parameters, and do proper constructor chaining (see comment above)

elludraon (approved):
Looks good. 2 minor style comments then feel free to push.

src/main/java/com/facebook/openwifirrm/optimizers/channel/RandomChannelInitializer.java
RockyMandayam2: Good catch, agreed

PR #48, opened 2022-08-25
[WIFI-10655] Gracefully abort topology sync with prov in case of bad data

jwoos:
Provisioning service had a bug where it allowed duplicate devices across different venues. RRM has the functionality to check for it but upon receiving this bad data, it will just crash the whole process. Handle it gracefully by handling `IllegalArgumentException`, log the error, and just abort the sync.

Open question:
- How do I test? I thought I'd go about mocking it but looking at the testing framework used (Junit), mocking is not included. Should I bring in Mockito, mock out the client and test?

### DIFF
--- a/src/main/java/com/facebook/openwifirrm/modules/ProvMonitor.java
+++ b/src/main/java/com/facebook/openwifirrm/modules/ProvMonitor.java
@@ -142,7 +142,14 @@ public class ProvMonitor implements Runnable {
 				topo.computeIfAbsent(venue, k -> new TreeSet<>());
 			zone.add(tag.serialNumber);
 		}
-		deviceDataManager.setTopology(topo);
+
+		try {
+			deviceDataManager.setTopology(topo);
+		} catch (IllegalArgumentException exc) {
+			logger.info("Failed validating the topology ({}), aborting sync", exc.getMessage());
+			return;
+		}
+
 		logger.info(
 			"Synced topology with owprov: {} zone(s), {} total device(s)",
 			topo.size(),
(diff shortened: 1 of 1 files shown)


src/main/java/com/facebook/openwifirrm/modules/ProvMonitor.java
```diff
@@ -142,7 +142,14 @@ protected void syncDataToProv(
 				topo.computeIfAbsent(venue, k -> new TreeSet<>());
 			zone.add(tag.serialNumber);
 		}
-		deviceDataManager.setTopology(topo);
+
+		try {
+			deviceDataManager.setTopology(topo);
+		} catch (IllegalArgumentException exc) {
```
elludraon, line 148: style: this is almost always named `e` (or `e1`, `e2`, etc. if that is already taken)

src/main/java/com/facebook/openwifirrm/modules/ProvMonitor.java
```diff
@@ -142,7 +142,14 @@ protected void syncDataToProv(
...
+			logger.info("Failed validating the topology ({}), aborting sync", exc.getMessage());
```
elludraon, line 149: use `logger.error`, like so:
```java
logger.error("Invalid topology received from owprov, aborting sync", e);
```

elludraon (changes requested):

elludraon:
You don't need to add unit tests for this, the `ProvMonitor` unit tests are still TBD since the Prov sync is not finalized yet (we're waiting on new config schemas to be implemented on their end).

elludraon (approved):

PR #49, opened 2022-08-25
Add autoformatter (Spotless with Maven, Eclipse JDT)

elludraon:
Add Java code auto-formatter config using Spotless (Maven) with the Eclipse JDT formatter. Add a custom Eclipse style XML which closely matches the existing (manual) code style.

To run the auto-formatter, use `mvn spotless:apply`. **This PR does not apply code changes. They will be applied later.**

Notable style differences (mainly due to formatter limitations):
- 80 chars per line will be enforced (except in some odd cases)
- Spacing around binary operators will be enforced (mostly)
- Even if 2+ arguments fit on a single overflow line, they will still be split 1-per-line
- Array initializers will be split 1-per-line
- 1 space preceding inline comment (previously 2 spaces)
- Overflow indentation will be applied in condition expressions

Notable formatter settings:
- Comment formatting is DISABLED (required too many manual fixes, maybe we can enable this later)
- Off/On Tags are ENABLED via default `@formatter:off` and `@formatter:on` comments (some are added in this PR)
- "Never join already wrapped lines" is ENABLED to prevent botching already-readable code

Also use Spotless to enforce copyright headers.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,6 +11,8 @@
     <swagger.version>2.1.10</swagger.version>
     <mainClassName>com.facebook.openwifirrm.Launcher</mainClassName>
     <appendVersionString></appendVersionString>
+    <!-- do not abort builds on autoformatter errors -->
+    <spotless.check.skip>true</spotless.check.skip>
   </properties>
   <build>
     <finalName>openwifi-rrm</finalName>
@@ -110,6 +112,32 @@
             <phase>compile</phase>
             <goals>
               <goal>resolve</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
+        <groupId>com.diffplug.spotless</groupId>
+        <artifactId>spotless-maven-plugin</artifactId>
+        <version>2.25.0</version>
+        <configuration>
+          <java>
+            <eclipse>
+              <file>${project.basedir}/spotless/eclipse-java-formatter.xml</file>
+              <version>4.12.0</version>
+            </eclipse>
+            <trimTrailingWhitespace />
+            <removeUnusedImports/>
+            <licenseHeader>
+              <file>${project.basedir}/spotless/license-header.txt</file>
+            </licenseHeader>
+          </java>
+        </configuration>
+        <executions>
+          <execution>
+            <phase>verify</phase>
+            <goals>
+              <goal>check</goal>
             </goals>
           </execution>
         </executions>
--- a/spotless/eclipse-java-formatter.xml
+++ b/spotless/eclipse-java-formatter.xml
@@ -1,0 +1,354 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<profiles version="16">
+    <profile kind="CodeFormatterProfile" name="Spotless" version="16">
+        <setting id="org.eclipse.jdt.core.formatter.insert_space_after_ellipsis" value="insert"/>
+        <setting id="org.eclipse.jdt.core.formatter.insert_space_after_comma_in_enum_declarations" value="insert"/>
+        <setting id="org.eclipse.jdt.core.formatter.insert_space_before_comma_in_allocation_expression" value="do not insert"/>
+        <setting id="org.eclipse.jdt.core.formatter.insert_space_before_at_in_annotation_type_declaration" value="insert"/>
+        <setting id="org.eclipse.jdt.core.formatter.parentheses_positions_in_for_statment" value="separate_lines_if_wrapped"/>
+        <setting id="org.eclipse.jdt.core.formatter.comment.new_lines_at_block_boundaries" value="true"/>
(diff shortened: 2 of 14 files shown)


jwoos (approved):
Looks good! I think having a linter is great :)

wuweicaifb:
I found an intellij plugin named [Spotless Gradle](https://plugins.jetbrains.com/plugin/18321-spotless-gradle). Do you know if this plugin share the same formatting rules?

pom.xml
```diff
@@ -11,6 +11,8 @@
     <swagger.version>2.1.10</swagger.version>
     <mainClassName>com.facebook.openwifirrm.Launcher</mainClassName>
     <appendVersionString></appendVersionString>
+    <!-- do not abort builds on autoformatter errors -->
```
RockyMandayam2, line 14: Good thought/catch!

spotless/eclipse-java-formatter.xml
```diff
@@ -0,0 +1,354 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<profiles version="16">
+    <profile kind="CodeFormatterProfile" name="Spotless" version="16">
```
RockyMandayam2, line 3: This probably doesn't fall under "formatting" but is there a way to add `@Override` for all overriding methods? I.e., when the compiler sees a method that is overriding  a parent method but is not marked with `@Override`, I'd like to auto-add `@Override` on the previous line.

RockyMandayam2, line 3: This probably doesn't fall under "formatting" but is there a way to remove `throws Exception` if the method does not actually throw an exception?

spotless/eclipse-java-formatter.xml
```diff
@@ -0,0 +1,354 @@
...
+        <setting id="org.eclipse.jdt.core.formatter.parentheses_positions_in_method_invocation" value="separate_lines_if_wrapped"/>
+        <setting id="org.eclipse.jdt.core.formatter.insert_space_between_empty_parens_in_enum_constant" value="do not insert"/>
+        <setting id="org.eclipse.jdt.core.formatter.blank_lines_after_imports" value="1"/>
+        <setting id="org.eclipse.jdt.core.formatter.insert_space_before_closing_paren_in_while" value="do not insert"/>
+        <setting id="org.eclipse.jdt.core.formatter.comment.insert_new_line_before_root_tags" value="insert"/>
+        <setting id="org.eclipse.jdt.core.formatter.insert_space_between_empty_parens_in_annotation_type_member_declaration" value="do not insert"/>
+        <setting id="org.eclipse.jdt.core.formatter.insert_space_before_comma_in_method_declaration_throws" value="do not insert"/>
+        <setting id="org.eclipse.jdt.core.formatter.parentheses_positions_in_switch_statement" value="separate_lines_if_wrapped"/>
+        <setting id="org.eclipse.jdt.core.formatter.comment.format_javadoc_comments" value="false"/>
```
RockyMandayam2, line 22: Just personal preference, but I like javadoc formatting (assuming this is what I think it is), just leaving my thoughts, no action needed.

spotless/eclipse-java-formatter.xml
RockyMandayam2, line 3: This may not be formatting, but do we have any thoughts on using `this.` when it's not needed? Also, I apologize if some of this stuff is in the xml and I just missed it

RockyMandayam2, line 3: Can we enforce that `public static final` (maybe all `static final`? not sure) vars are in ALL_CAPS_WITH_UNDERSCORES format?

RockyMandayam2, line 3: Is there a way to remove unnecessary casts?

RockyMandayam2 (approved):
All my comments are questions/preferences, no blocker

spotless/eclipse-java-formatter.xml
elludraon: I felt that Eclipse does too much on comment formatting that is unable to be disabled, particularly around removal of whitespace that is there for readability.

Feel free to experiment with it afterwards but it clobbered way too much of the existing code for me to consider this.

elludraon (commented):
@RockyMandayam2 - Almost all of your comments (besides the comment formatting) are not formatter settings and can be separately configured in your IDE. Formatter is also not a linter (which could enforce variable naming conventions, etc.), such as checkstyle.

elludraon:
> I found an intellij plugin named [Spotless Gradle](https://plugins.jetbrains.com/plugin/18321-spotless-gradle). Do you know if this plugin share the same formatting rules?

@wuweicaifb - Yeah, I saw that too. Gradle is a different build system though. Perhaps you can see if that plugin still works if you use a dummy `build.gradle` only holding the Spotless config?

PR #50, opened 2022-08-26
Refactor tests

RockyMandayam2:
This updates TestUtils, focusing on the methods that create state, to help testing in future PRs that will test TPC algorithms in a multi-band multi-radio setting:
- for the `createState` that was supposed to create a state with only one radio, make it actually only do one radio (it did two before)
- for the `createState` that was supposed to create a state with two radios, make it also create an interface for each of these radios (there was only one such interface before)
- update javadocs


src/test/java/com/facebook/openwifirrm/optimizers/TestUtils.java
```diff
@@ -263,16 +257,136 @@ public static State createState(
...
+	 * 
+	 * @param channel channel number
+	 * @param channelWidth channel width in MHz
+	 * @param bssid bssid
+	 * @return the state of an AP with one radio
+	 */
+	public static State createState(int channel, int channelWidth, String bssid) {
+		final int numRadios = 1;
+		String ifaceString = createStateInterfaceString(0);
+		StringBuffer stateString = new StringBuffer();
```
elludraon, line 290: nit: this is explicitly not a string, so don't name it like this - usually `StringBuffer` var is named `sb`

elludraon (commented):
The point of leaving the raw JSON (aside from being lazy) was so we could copy/paste a new blob if the schema changed. If you're dynamically building these JSON objects and duplicating them in different methods, you should probably just build them in code as `JSONObject`.

src/main/java/com/facebook/openwifirrm/optimizers/channel/LeastUsedChannelOptimizer.java
```diff
@@ -363,6 +363,8 @@ public Map<String, Map<String, Integer>> computeChannelMap() {
 				int[] currentChannelInfo = getCurrentChannel(band, serialNumber, state);
 				int currentChannel = currentChannelInfo[0];
 				int currentChannelWidth = currentChannelInfo[1];
+				// TODO Look into whether we should filter as described below
+				// TODO Look into if we actually filter...
```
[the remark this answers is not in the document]
RockyMandayam2: Ok, I will make a note to myself and if needed do another PR down the road.

src/test/java/com/facebook/openwifirrm/optimizers/TestUtils.java
```diff
@@ -204,113 +216,128 @@ public static List<WifiScanEntry> createWifiScanList(
...
-	/** Create a device state object with the given radio channel. */
-	public static State createState(int channel, int channelWidth, String bssid) {
-		return createState(channel, channelWidth, 20, 1, 20, 0, bssid);
+	/**
+	 * Create an uplink interface with one radio, to place in
+	 * {@link State#interfaces}.
+	 *
+	 * @param index index of this interface in {@link State#interfaces}
+	 * @return an uplink interface with no radios, to place in
+	 *         {@link State#interfaces}
```
elludraon, line 225: nit: you don't need to `@link` so many times repeatedly

src/test/java/com/facebook/openwifirrm/optimizers/TestUtils.java
```diff
@@ -204,113 +216,128 @@ public static List<WifiScanEntry> createWifiScanList(
...
+	 */
+	private static State.Interface createUp0StateInterface(int index) {
```
elludraon, line 227: Can you remove the `0` and `1` in these method names? It makes too little sense to someone reading it, just put the hardcoded name in the javadoc comment.

src/test/java/com/facebook/openwifirrm/optimizers/TestUtils.java
```diff
@@ -322,18 +349,96 @@ public static State createState(
...
+	public static State createState(int channel, int channelWidth, String bssid) {
+		return createState(
+			new int[] {channel},
+			new int[] {channelWidth},
+			new String[] {bssid}
+		);
+	}
+
+	/**
+	 * Create a device state object with one radio.
```
elludraon, line 416: two radios?

elludraon (approved):
That's a lot better.

PR #51, opened 2022-08-26
Enable MeasurementBasedApClientTPC in 2G

wuweicaifb:
1. Add a new util method to determine the band with the given channel.
2. Instead of only using the first radio, iterate over all the radios for each device when compute Tx power in `MeasurementBasedApClientTPC`.
3. Instead of hardcode the band as `5G`, leverage the new method to determine the band for each radio.
4. Add test cases for the changes.

@pohanhf some questions regarding the radios and channels:
1. Is it possible to have a channel that cannot be mapped to either 2G or 5G?
2. Is it possible to have a device with two channels mapping to the same band? e.g. A device with two radios, whose channels are 36 and 60.

### DIFF
--- a/src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApClientTPC.java
+++ b/src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApClientTPC.java
@@ -108,7 +108,7 @@ public class MeasurementBasedApClientTPC extends TPC {
 		return SNR_min + currentTxPower - clientRssi + NP + NF - M;
 	}
 
-	/** Compute adjusted tx power (dBm) for the given radio. */
+	/** Compute new tx power (dBm) for the given radio. */
 	private int computeTxPowerForRadio(
 		String serialNumber, State state, JsonObject radio
 	) {
--- a/src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApClientTPC.java
+++ b/src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApClientTPC.java
@@ -249,6 +249,8 @@ public class MeasurementBasedApClientTPC extends TPC {
 				if (currentChannel == null) {
 					continue;
 				}
+
+				// TODO: check if band can be directly got from config rather than mapped from channel
 				String band = UCentralUtils.getBandFromChannel(currentChannel);
 				if (band == null) {
 					continue;
--- a/src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApClientTPC.java
+++ b/src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApClientTPC.java
@@ -250,7 +250,6 @@ public class MeasurementBasedApClientTPC extends TPC {
 					continue;
 				}
 
-				// TODO: check if band can be directly got from config rather than mapped from channel
 				String band = UCentralUtils.getBandFromChannel(currentChannel);
 				if (band == null) {
 					continue;
--- a/src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApClientTPC.java
+++ b/src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApClientTPC.java
@@ -15,6 +15,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.TreeMap;
 
+import com.facebook.openwifirrm.ucentral.UCentralUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -107,12 +108,11 @@ public class MeasurementBasedApClientTPC extends TPC {
 		return SNR_min + currentTxPower - clientRssi + NP + NF - M;
 	}
 
-	/** Compute adjusted tx power (dBm) for the given device. */
-	private int computeTxPowerForDevice(
-		String serialNumber, State state, int radioIndex
+	/** Compute adjusted tx power (dBm) for the given radio. */
+	private int computeTxPowerForRadio(
+		String serialNumber, State state, JsonObject radio
 	) {
 		// Find current tx power and bandwidth
-		JsonObject radio = state.radios[radioIndex];
 		int currentTxPower =
 			radio.has("tx_power") && !radio.get("tx_power").isJsonNull()
 				? radio.get("tx_power").getAsInt()
@@ -239,13 +239,25 @@ public class MeasurementBasedApClientTPC extends TPC {
 				);
 				continue;
 			}
-			// TODO: only using first radio
-			final int radioIndex = 0;
-
-			int newTxPower =
-				computeTxPowerForDevice(serialNumber, state, radioIndex);
 			Map<String, Integer> radioMap = new TreeMap<>();
-			radioMap.put(UCentralConstants.BAND_5G, newTxPower);
+
+			for (JsonObject radio: state.radios) {
+				Integer currentChannel =
+					radio.has("channel") && !radio.get("channel").isJsonNull()
+						? radio.get("channel").getAsInt()
+						: null;
+				if (currentChannel == null) {
+					continue;
+				}
+				String band = UCentralUtils.getBandFromChannel(currentChannel);
+				if (band == null) {
+					continue;
+				}
+				int newTxPower =
+					computeTxPowerForRadio(serialNumber, state, radio);
+
+				radioMap.put(band, newTxPower);
+			}
 			txPowerMap.put(serialNumber, radioMap);
 		}
 
--- a/src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
+++ b/src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
@@ -391,7 +391,6 @@ public class UCentralUtils {
 				&& channel <= UPPER_CHANNEL_LIMIT.get(band);
 	}
 
-	// TODO: Remove this method if band can be directly got from config
 	/**
 	 * Given the channel, gets the band by checking lower bound and upper bound
 	 * of each band
--- a/src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
+++ b/src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
@@ -390,4 +390,13 @@ public class UCentralUtils {
 		return LOWER_CHANNEL_LIMIT.get(band) <= channel
 				&& channel <= UPPER_CHANNEL_LIMIT.get(band);
 	}
+
+	public static String getBandFromChannel(int channel) {
+		for (String band: UCentralConstants.BANDS) {
+			if (isChannelInBand(channel, band)) {
+				return band;
+			}
+		}
+		return null;
+	}
 }
--- a/src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
+++ b/src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
@@ -391,6 +391,7 @@ public class UCentralUtils {
 				&& channel <= UPPER_CHANNEL_LIMIT.get(band);
 	}
 
+	// TODO: Remove this method if band can be directly got from config
 	/**
 	 * Given the channel, gets the band by checking lower bound and upper bound
 	 * of each band
--- a/src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
+++ b/src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
@@ -391,6 +391,13 @@ public class UCentralUtils {
 				&& channel <= UPPER_CHANNEL_LIMIT.get(band);
 	}
 
+	/**
+	 * Given the channel, gets the band by checking lower bound and upper bound
+	 * of each band
+	 *
+	 * @param channel channel number
+	 * @return band if the channel can be mapped to a valid band; null otherwise
+	 */
 	public static String getBandFromChannel(int channel) {
 		for (String band: UCentralConstants.BANDS) {
 			if (isChannelInBand(channel, band)) {
--- a/src/test/java/com/facebook/openwifirrm/optimizers/TestUtils.java
+++ b/src/test/java/com/facebook/openwifirrm/optimizers/TestUtils.java
@@ -418,6 +418,8 @@ public class TestUtils {
 			state.interfaces[i].ssids[0].associations =
 				new State.Interface.SSID.Association[clientRssis[i].length];
 			for (int j = 0; j < clientRssis[i].length; j++) {
+				state.interfaces[i].ssids[0].associations[j] =
+					state.interfaces[i].ssids[0].new Association();
 				state.interfaces[i].ssids[0].associations[j].rssi = clientRssis[i][j];
 			}
 		}
@@ -453,11 +455,7 @@ public class TestUtils {
 	}
 
 	/**
-<<<<<<< HEAD
 	 * Create a device state object with one radio.
-=======
-	 * Create a device state object with one radio and client RSSIs.
->>>>>>> 37806df (Resolve comments)
 	 *
 	 * @param channel channel number
 	 * @param channelWidth channel width in MHz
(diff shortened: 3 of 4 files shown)


wuweicaifb:
@elludraon When I ran "mvn spotless:apply", it formats all the files in the repository. Any way to only work on the changed files? I guess you will have a PR to format them, so I will revert the latest commit.

src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApClientTPC.java
```diff
@@ -239,13 +239,25 @@ public Map<String, Map<String, Integer>> computeTxPowerMap() {
...
+
+			for (JsonObject radio: state.radios) {
+				Integer currentChannel =
+					radio.has("channel") && !radio.get("channel").isJsonNull()
+						? radio.get("channel").getAsInt()
+						: null;
+				if (currentChannel == null) {
+					continue;
+				}
+				String band = UCentralUtils.getBandFromChannel(currentChannel);
```
elludraon, line 252: Doesn't this object already have a "band" field you can use directly?

elludraon (changes requested):

src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApClientTPC.java
RockyMandayam2: No, because this is from the state which only has channel. The device status has band but not bssid.

elludraon:
> @elludraon When I ran "mvn spotless:apply", it formats all the files in the repository. Any way to only work on the changed files? I guess you will have a PR to format them, so I will revert the latest commit.

@wuweicaifb - If you want to apply the formatting changes for only your modified files, I believe you can use "ratchetFrom" in your `pom.xml` plugin configuration block:
```
<plugin>
  <groupId>com.diffplug.spotless</groupId>
  <artifactId>spotless-maven-plugin</artifactId>
  ...
  <configuration>
    <ratchetFrom>origin/main</ratchetFrom>  <!-- THIS LINE -->
    ...
  </configuration>
</plugin>
```

RockyMandayam2:
> 1. Add a new util method to determine the band with the given channel.
> 2. Instead of only using the first radio, iterate over all the radios for each device when compute Tx power in `MeasurementBasedApClientTPC`.
> 3. Instead of hardcode the band as `5G`, leverage the new method to determine the band for each radio.
> 4. Add test cases for the changes.
> 
> @pohanhf some questions regarding the radios and channels:
> 
> 1. Is it possible to have a channel that cannot be mapped to either 2G or 5G?
> 2. Is it possible to have a device with two channels mapping to the same band? e.g. A device with two radios, whose channels are 36 and 60.

I'm not Po-Han but I know about the first one:
1. We are only dealing with 2G and 5G now, but yes there may be 6G but also there are some "edge cases" of sub-1G bands and all... I'd say if it's outside of 2G and 5G, you can mostly ignore it for now... but please write the code so that we can add 6G to this list soon :)

src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApClientTPC.java
elludraon: OK - would still be a lot better to read the band from device config - do we know if the array indexes are correlated?

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
```diff
@@ -390,4 +390,20 @@ public static boolean isChannelInBand(int channel, String band) {
...
 	}
+
+	/**
+	 * Given the channel, gets the band by checking lower bound and upper bound
+	 * of each band
+	 *
+	 * @param channel channel number
+	 * @return band if the channel can be mapped to a valid band; null otherwise
+	 */
+	public static String getBandFromChannel(int channel) {
```
RockyMandayam2, line 401: An alternate suggestion below, I'm not actually sure which is clearer/better practice. @elludraon thoughts?

`UCentralConstants.BANDS.stream().filter(b -> isChannelInBand(channel, b)).findAny().orElse(null);`

src/test/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApClientTPCTest.java
```diff
@@ -31,15 +33,21 @@ public class MeasurementBasedApClientTPCTest {
 
 	/** Create a device state object containing the given parameters. */
 	private State createState(
```
RockyMandayam2, line 35: I have another PR editing some of this stuff. Let me look into what you did and we can discuss how to resolve on Mon.

src/test/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApClientTPCTest.java
```diff
@@ -114,4 +122,56 @@ void test1() throws Exception {
...
+		dataModel.latestState.put(
+			deviceD,
+			createState(deviceA, 20 /*txPower*/, 20 /*bandwidth*/, new int[]{25} /*channel*/)
+		);
+
+		TPC optimizer = new MeasurementBasedApClientTPC(dataModel, TEST_ZONE, deviceDataManager);
+		Map<String, Map<String, Integer>> txPowerMap =
+			optimizer.computeTxPowerMap();
+
+		Set<String> expectedBands = new TreeSet<>();
```
RockyMandayam2, line 167: We want to test more than just which bands are present I think. We want to test that the alg actually runs correctly independently on each band. If that doesn't make sense, let me know I can explain more.

RockyMandayam2 (changes requested):

src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApClientTPC.java
RockyMandayam2: They may be, but I don't know how to confirm it. Also, why is it better to read from device config?

src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApClientTPC.java
```diff
@@ -107,12 +108,11 @@ private double computeTxPower(
 		return SNR_min + currentTxPower - clientRssi + NP + NF - M;
```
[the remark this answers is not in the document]
elludraon: These are copied from the design doc

src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApClientTPC.java
elludraon: Why would you indirectly infer the band based on another field when the band is explicitly written?

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
elludraon: This function should not exist (convo above)

pohanhf:
Q: Is it possible to have a channel that cannot be mapped to either 2G or 5G?
A: No, channels 1 to 11 are only for 2G. channels 36 to 165 are only for 5G. Also, when we talk about channel here, we only mean "primary" channel. In other words, channel 38 is not the case here. 

Q: Is it possible to have a device with two channels mapping to the same band? e.g. A device with two radios, whose channels are 36 and 60.
A: No, it is not possible as the above explanation. 

Let me know if there are any questions. :)

RockyMandayam2:
> Q: Is it possible to have a channel that cannot be mapped to either 2G or 5G? A: No, channels 1 to 11 are only for 2G. channels 36 to 165 are only for 5G. Also, when we talk about channel here, we only mean "primary" channel. In other words, channel 38 is not the case here.
> 
> Q: Is it possible to have a device with two channels mapping to the same band? e.g. A device with two radios, whose channels are 36 and 60. A: No, it is not possible as the above explanation.
> 
> Let me know if there are any questions. :)

Po-Han, I don't think the first answers the second. The second is asking if an AP can be part of two different channels on the same band... E.g., maybe one AP is used for two SSIDs on different channels and both happen to be 5G. Is that allowed?

pohanhf:
> > Q: Is it possible to have a channel that cannot be mapped to either 2G or 5G? A: No, channels 1 to 11 are only for 2G. channels 36 to 165 are only for 5G. Also, when we talk about channel here, we only mean "primary" channel. In other words, channel 38 is not the case here.
> > Q: Is it possible to have a device with two channels mapping to the same band? e.g. A device with two radios, whose channels are 36 and 60. A: No, it is not possible as the above explanation.
> > Let me know if there are any questions. :)
> 
> Po-Han, I don't think the first answers the second. The second is asking if an AP can be part of two different channels on the same band... E.g., maybe one AP is used for two SSIDs on different channels and both happen to be 5G. Is that allowed?

Yes, it is allowed. It is called dual 5GHz radio mode (or something similar). In that case, that AP should have two unique bssids. Using bssids should be the way to do this.

src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApClientTPC.java
RockyMandayam2: Makes sense.

I guess @wuweicaifb can look in uCentral repo to see how they generate the device status (which they call config I think) vs how they generate the state (which they call stats (I showed you an example, I think called "nat_stats.json" or something)). Maybe you can tell if the elements are the same or correlated in some way.

The caveat is that if they change this later, we could be in trouble... since there is no explicit definition api/yaml as far as I know.

src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApClientTPC.java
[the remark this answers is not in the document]
RockyMandayam2: Ok. Up to Wuwei/Jeffrey what to do

wuweicaifb:
Deleted the branch by mistake. Restore the branch and reopen the PR now.

wuweicaifb:
> > @elludraon When I ran "mvn spotless:apply", it formats all the files in the repository. Any way to only work on the changed files? I guess you will have a PR to format them, so I will revert the latest commit.
> 
> @wuweicaifb - If you want to apply the formatting changes for only your modified files, I believe you can use "ratchetFrom" in your `pom.xml` plugin configuration block:
> 
> ```
> <plugin>
>   <groupId>com.diffplug.spotless</groupId>
>   <artifactId>spotless-maven-plugin</artifactId>
>   ...
>   <configuration>
>     <ratchetFrom>origin/main</ratchetFrom>  
>     ...
>   </configuration>
> </plugin>
> ```

It's just kinda inconvenient to adjust the xml file manually every time. Is there a way to use a parameter in the CLI command to choose files?

src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApClientTPC.java
wuweicaifb: Is `DeviceWithStatus` the device config you mentioned? It has no `band` as an attribute. I think it's safe here to infer the band from channel as long as the data is consistent, but I agree if the band is explicitly provided, directly reading is better. I will file a separate task to investigate how we could add band in a decent way and do it in a new PR.

src/test/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApClientTPCTest.java
wuweicaifb: "the alg actually runs correctly independently" is tested in `testComputeCorrectTxPower`

elludraon:
@wuweicaifb - Just waiting for most people to push their PRs first so everyone doesn't get merge conflicts. Will apply it to all files.

src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApClientTPC.java
elludraon: Yes, it's in `DeviceWithStatus.configuration.radios[n].band`. Can you add a TODO comment in this code here as well as `getBandFromChannel()` to remove it later?

src/test/java/com/facebook/openwifirrm/optimizers/TestUtils.java
```diff
@@ -408,10 +417,37 @@ public static State createState(int channel, int channelWidth, String bssid) {
...
 			new int[] {channelWidth},
-			new String[] {bssid}
+			new int[] {DEFAULT_CHANNEL_WIDTH},
+			new String[] {bssid},
+			new int[][] {{}}
 		);
 	}
 
+	/**
+	 * Create a device state object with one radio and client Rssis.
```
elludraon, line 427: nit: all caps "RSSI"

elludraon (approved):

src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApClientTPC.java
elludraon: Discussed this with @RockyMandayam2 and we confirmed they are currently NOT correlated. So no need for TODO, this can stay. Sorry for the thrash.

RockyMandayam2 (approved):
PR approved. Minor thing - maybe you can rename `testVariousBands` to `testMultipleBands`.

src/test/java/com/facebook/openwifirrm/optimizers/TestUtils.java
```diff
@@ -369,7 +369,7 @@ private static State.Unit createStateUnit() {
 	 * @param bssids array of BSSIDs
 	 * @return the state of an AP with radios described by the given parameters
 	 */
-	public static State createState(int[] channels, int[] channelWidths, String[] bssids) {
+	public static State createState(int[] channels, int[] channelWidths, int[] txPowers, String[] bssids, int[][] clientRssis) {
```
elludraon, line 372: please add `@param clientRssis` into the javadoc above (since all other params are documented already)

src/test/java/com/facebook/openwifirrm/optimizers/TestUtils.java
```diff
@@ -408,10 +417,37 @@ public static State createState(int channel, int channelWidth, String bssid) {
...
+			new int[][] {{}}
 		);
 	}
 
+	/**
+	 * Create a device state object with one radio and client RSSIs.
+	 *
+	 * @param channel channel number
+	 * @param channelWidth channel width in MHz
+	 * @param bssid bssid
```
elludraon, line 431: same here, if you are going to document any parameters with `@param` you should document all of them

src/test/java/com/facebook/openwifirrm/optimizers/TestUtils.java
```diff
@@ -438,7 +474,43 @@ public static State createState(
...
+	 *
+	 * @param channelA channel number
+	 * @param channelWidthA channel width (MHz) of channelA
+	 * @param txPowerA tx power for channelA
+	 * @param bssidA bssid for radio on channelA
+	 * @param channelB channel number
+	 * @param channelWidthB channel width (MHz) of channelB
+	 * @param txPowerB tx power for channelB
+	 * @param bssidB bssid for radio on channelB
+	 * @return the state of an AP with two radios
```
elludraon, line 494: same here

elludraon (approved):
Looks good. Please address javadoc-related comments before pushing.

src/test/java/com/facebook/openwifirrm/optimizers/TestUtils.java
```diff
@@ -437,11 +447,38 @@ public static State createState(int channel, int channelWidth, String bssid) {
...
+			channel, channelWidth, txPower, bssid, new int[] {}
+		);
+	}
+
+	/**
+<<<<<<< HEAD
+	 * Create a device state object with one radio.
+=======
+	 * Create a device state object with one radio and client RSSIs.
+>>>>>>> 37806df (Resolve comments)
```
elludraon, line 460: leftover rebase conflict

PR #52, opened 2022-08-26
Set thread names

elludraon:
Set thread names for all program threads (modules and executors). `NamedThreadFactory` is a copy of `Executors.DefaultThreadFactory` which accepts a name prefix.

### DIFF
--- a/src/main/java/com/facebook/openwifirrm/Utils.java
+++ b/src/main/java/com/facebook/openwifirrm/Utils.java
@@ -19,6 +19,8 @@ import java.nio.file.Files;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Scanner;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import org.json.JSONObject;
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/facebook/openwifirrm/Utils.java
```diff
@@ -57,6 +59,40 @@ protected boolean removeEldestEntry(final Map.Entry<K, V> eldest) {
 		}
 	}
 
+	/** Thread factory based on DefaultThreadFactory supporting name prefix. */
+	public static class NamedThreadFactory implements ThreadFactory {
+		private static final AtomicInteger poolNumber = new AtomicInteger(1);
```
RockyMandayam2, line 64: Maybe add javadocs to all these member vars

RockyMandayam2 (approved):

src/main/java/com/facebook/openwifirrm/Utils.java
elludraon: This was just copy/pasted from the JDK sources, which is why I didn't bother.

PR #54, opened 2022-08-27
Add multi-band check for MeasurementBasedApApTPC

RockyMandayam2:
In `MeasurementBasedApApTPC`, make sure it only sets tx power  for a (device, band) if that device operates in that band. This PR includes:
- A few lines in `MeasurementBasedApApTPC` to implement this behavior
- Many changes in `MeasurementBasedApApTPCTest` to test the behavior
- Changes to `TestUtils` both to support the changes in `MeasurementBasedApApTPCTest` and to fix a minor bug from a previous PR

### DIFF
--- a/src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPC.java
+++ b/src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPC.java
@@ -267,6 +267,10 @@ public class MeasurementBasedApApTPC extends TPC {
 			}
 			JsonArray radioStatuses = allStatuses.get(serialNumber).getAsJsonArray();
 			int currentTxPower = getCurrentTxPower(radioStatuses, band);
+			if (currentTxPower == 0) {
+				// this AP is not on the band of interest
+				continue;
+			}
 			String bssid = state.interfaces[0].ssids[0].bssid;
 			List<Integer> rssiValues = bssidToRssiValues.get(bssid);
 			logger.debug("Device <{}> : BSSID <{}>", serialNumber, bssid);
--- a/src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPC.java
+++ b/src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPC.java
@@ -14,6 +14,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.TreeMap;
 
@@ -112,14 +113,17 @@ public class MeasurementBasedApApTPC extends TPC {
 	}
 
 	/**
-	 * Get the current band radio tx power (the first one found) for an AP using the
-	 * latest device status.
+	 * Get the current band radio tx power (the first one found) for an AP using
+	 * the latest device status.
 	 *
 	 * @param latestDeviceStatus JsonArray containing radio config for the AP
-	 * @param band "2G" or "5G"
-	 * @return the tx power, or -1 if none found
+	 * @param band band (e.g., "2G")
+	 * @return an Optional containing the tx power if one exists, or else an
+	 *         empty Optional
 	 */
-	protected static int getCurrentTxPower(JsonArray latestDeviceStatus, String band) {
+	protected static Optional<Integer> getCurrentTxPower(
+		JsonArray latestDeviceStatus, String band
+	) {
 		for (JsonElement e : latestDeviceStatus) {
 			if (!e.isJsonObject()) {
 				continue;
@@ -127,10 +131,10 @@ public class MeasurementBasedApApTPC extends TPC {
 			JsonObject radioObject = e.getAsJsonObject();
 			String radioBand = radioObject.get("band").getAsString();
 			if (radioBand.equals(band) && radioObject.has("tx-power")) {
-				return radioObject.get("tx-power").getAsInt();
+				return Optional.of(radioObject.get("tx-power").getAsInt());
 			}
 		}
-		return -1;
+		return Optional.empty();
 	}
 
 	/**
@@ -266,11 +270,14 @@ public class MeasurementBasedApApTPC extends TPC {
 				continue;
 			}
 			JsonArray radioStatuses = allStatuses.get(serialNumber).getAsJsonArray();
-			int currentTxPower = getCurrentTxPower(radioStatuses, band);
-			if (currentTxPower < 0) {
+			Optional<Integer> possibleCurrentTxPower = getCurrentTxPower(
+				radioStatuses, band
+			);
+			if (possibleCurrentTxPower.isEmpty()) {
 				// this AP is not on the band of interest
 				continue;
 			}
+			int currentTxPower = possibleCurrentTxPower.get();
 			String bssid = state.interfaces[0].ssids[0].bssid;
 			List<Integer> rssiValues = bssidToRssiValues.get(bssid);
 			logger.debug("Device <{}> : BSSID <{}>", serialNumber, bssid);
--- a/src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPC.java
+++ b/src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPC.java
@@ -117,7 +117,7 @@ public class MeasurementBasedApApTPC extends TPC {
 	 *
 	 * @param latestDeviceStatus JsonArray containing radio config for the AP
 	 * @param band "2G" or "5G"
-	 * @return the tx power, or 0 if none found
+	 * @return the tx power, or -1 if none found
 	 */
 	protected static int getCurrentTxPower(JsonArray latestDeviceStatus, String band) {
 		for (JsonElement e : latestDeviceStatus) {
@@ -130,7 +130,7 @@ public class MeasurementBasedApApTPC extends TPC {
 				return radioObject.get("tx-power").getAsInt();
 			}
 		}
-		return 0;
+		return -1;
 	}
 
 	/**
@@ -267,7 +267,7 @@ public class MeasurementBasedApApTPC extends TPC {
 			}
 			JsonArray radioStatuses = allStatuses.get(serialNumber).getAsJsonArray();
 			int currentTxPower = getCurrentTxPower(radioStatuses, band);
-			if (currentTxPower == 0) {
+			if (currentTxPower < 0) {
 				// this AP is not on the band of interest
 				continue;
 			}
--- a/src/test/java/com/facebook/openwifirrm/optimizers/TestUtils.java
+++ b/src/test/java/com/facebook/openwifirrm/optimizers/TestUtils.java
@@ -18,6 +18,7 @@ import java.util.TreeSet;
 import java.util.stream.Collectors;
 
 import com.facebook.openwifirrm.DeviceTopology;
+import com.facebook.openwifirrm.ucentral.UCentralConstants;
 import com.facebook.openwifirrm.ucentral.UCentralUtils;
 import com.facebook.openwifirrm.ucentral.UCentralUtils.WifiScanEntry;
 import com.facebook.openwifirrm.ucentral.models.State;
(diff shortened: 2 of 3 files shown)


src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPC.java
```diff
@@ -130,7 +130,7 @@ protected static int getCurrentTxPower(JsonArray latestDeviceStatus, String band
 				return radioObject.get("tx-power").getAsInt();
 			}
 		}
-		return 0;
```
RockyMandayam2, line 133: I made this change to differentiate an AP that can operate on the band but just happens to be transmitting at 0 dBm vs an AP that cannot operate on the band

src/test/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPCTest.java
```diff
@@ -74,26 +76,39 @@ private static DeviceDataManager createDeviceDataManager() {
...
-	 * Creates a data model with 3 devices.
-	 * All are at max_tx_power, which represents the first step in greedy TPC.
+	 * Creates a data model with 3 devices. All are at max_tx_power, which
+	 * represents the first step in greedy TPC.
+	 *
+	 * @param deviceCIn5G true if device C should operate in the 5G band
+	 * @return a data model
 	 */
-	private static DataModel createModel() {
+	private static DataModel createModel(boolean deviceCIn5G) {
```
elludraon, line 85: This is a very odd function signature and makes it hard to read/understand the tests which call this. For example you could have 3 parameters where you pass in the band name (constant) for each device. Or take no arguments and modify the result (`DataModel`) afterwards where needed.

src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPC.java
```diff
@@ -267,6 +267,10 @@ protected void buildTxPowerMapForBand(String band, Map<String, Map<String, Integ
 			}
 			JsonArray radioStatuses = allStatuses.get(serialNumber).getAsJsonArray();
 			int currentTxPower = getCurrentTxPower(radioStatuses, band);
+			if (currentTxPower < 0) {
+				// this AP is not on the band of interest
+				continue;
+			}
```
elludraon, line 273: Similar to my comment on PR #51 , I prefer you add an explicit check for band instead of inferring it from a seemingly unrelated function. Just add a function like:
```java
// in UCentralUtils
public static String getRadioBand(int index, JsonArray radioStatuses) { ... }
```

elludraon (changes requested):

src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPC.java
```diff
@@ -130,7 +130,7 @@ protected static int getCurrentTxPower(JsonArray latestDeviceStatus, String band
 				return radioObject.get("tx-power").getAsInt();
 			}
 		}
-		return 0;
+		return -1;
```
wuweicaifb, line 133: I would prefer `null` to a magic number like `-1` when checking if the function was able to give a feasible return value.

src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPC.java
elludraon: Discussed this elsewhere - as we can't correlate the two structures, ignore this comment.

Another question though, why are we reading tx power from device config/status instead of state/statistics?

RockyMandayam2: There is no good reason I think. I think this is the way it already was. In `MeasurementBasedApClientTPC` it reads it from state.

src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPC.java
RockyMandayam2: It returns an int, so null wouldn't work I think, but I can do an `Optional`

src/test/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPCTest.java
RockyMandayam2: I will modify the result afterwards in the only place where `false` is passed in and keep the `true`-passed-in behavior the default.

src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPC.java
RockyMandayam2: I think if we want to make this "consistent" it should be a separate PR

src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPC.java
wuweicaifb: I'm good with either `Optional` or `Integer`.

src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPC.java
wuweicaifb: Would you like to include this in your PR or my PR? I'm also good with a separate one.

src/test/java/com/facebook/openwifirrm/optimizers/TestUtils.java
```diff
@@ -34,6 +35,8 @@ public class TestUtils {
 
 	/** Default channel width in MHz */
 	public static final int DEFAULT_CHANNEL_WIDTH = 20;
+	/** Default tx power in dBm */
+	public static final int DEFAULT_TX_POWER = 20;
```
wuweicaifb, line 39: I believe we have some `DEFAULT_TX_POWER` somewhere else. It's good to keep them consistent if we can. Try reusing the existing default?

wuweicaifb: It depends on whether we want those "default" to be identical logically. If not, I'm ok with keep them separate.

src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPC.java
wuweicaifb: Ok. Since you will be on vacation, I can take care of it, and publish on PR after both our current PRs are merged.

src/test/java/com/facebook/openwifirrm/optimizers/TestUtils.java
```diff
@@ -73,7 +77,26 @@ private static JsonObject createDeviceStatusRadioObject(
...
+		));
+		return jsonList;
+	}
+
+	/**
+	 * Create the device status array.
+	 *
+	 * @param band band (e.g., "2G")
+	 * @param channel channel number
+	 * @param txPower tx power in dBm
```
wuweicaifb, line 90: nit: Line 384 uses () to specify the unit. It's better to keep the same style.

src/test/java/com/facebook/openwifirrm/optimizers/TestUtils.java
```diff
@@ -366,12 +381,16 @@ private static State.Unit createStateUnit() {
 	 *
 	 * @param channels array of channel numbers
 	 * @param channelWidths array of channel widths (MHz)
+	 * @param txPowers array of tx powers (dBm)
 	 * @param bssids array of BSSIDs
 	 * @return the state of an AP with radios described by the given parameters
 	 */
-	public static State createState(int[] channels, int[] channelWidths, String[] bssids) {
+	public static State createState(
+		int[] channels, int[] channelWidths, int[] txPowers, String[] bssids
```
wuweicaifb, line 389: just FYI: I will add clientRssis to the arg list in PR #51

RockyMandayam2: No problem, you can change what you like when/after you rebase :)

src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPC.java
RockyMandayam2: Ok. I'll let you (and Jeffrey as needed) coordinate it, thanks!

wuweicaifb (approved):

src/test/java/com/facebook/openwifirrm/optimizers/TestUtils.java
RockyMandayam2: Different parts of the code do it differently (even outside my code). If you want to clean this up, as we discussed, please do it in a separate PR

src/test/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPCTest.java
```diff
@@ -306,7 +331,54 @@ private static void testComputeTxPowerMapInBand(String band) {
...
+		Map<String, Map<String, Integer>> txPowerMap = optimizer.computeTxPowerMap();
+
+		// test 2G band
+		assertEquals(3, txPowerMap.size());
+		assertEquals(2, txPowerMap.get(DEVICE_A).get(UCentralConstants.BAND_2G));
+		assertEquals(15, txPowerMap.get(DEVICE_B).get(UCentralConstants.BAND_2G));
+		assertEquals(10, txPowerMap.get(DEVICE_C).get(UCentralConstants.BAND_2G));
+
+		// test 5G band
+		assertEquals(3, txPowerMap.size());
```
elludraon, line 378: duplicate (line 372), delete

src/test/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPCTest.java
```diff
@@ -306,7 +331,54 @@ private static void testComputeTxPowerMapInBand(String band) {
...
 	void test_computeTxPowerMap() throws Exception {
-		testComputeTxPowerMapInBand(UCentralConstants.BAND_5G);
-		testComputeTxPowerMapInBand(UCentralConstants.BAND_2G);
+		// test each band separately
+		for (String band : UCentralConstants.BANDS) {
+			testComputeTxPowerMapSimpleInOneBand(band);
+			testComputeTxPowerMapMissingDataInOneBand(band);
+		}
+
+		// test both bands simultaneously with different setups on each band
```
elludraon, line 340: It looks like you should be splitting this into multiple test methods?

src/test/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPCTest.java
```diff
@@ -84,8 +88,9 @@ private static DataModel createModel() {
 			MAX_TX_POWER, BSSID_A, 36, DEFAULT_CHANNEL_WIDTH, MAX_TX_POWER, BSSID_A);
 		State stateB = TestUtils.createState(1, DEFAULT_CHANNEL_WIDTH,
 			MAX_TX_POWER, BSSID_B, 36, DEFAULT_CHANNEL_WIDTH, MAX_TX_POWER, BSSID_B);
-		State stateC = TestUtils.createState(1, DEFAULT_CHANNEL_WIDTH,
-			MAX_TX_POWER, BSSID_C, 36, DEFAULT_CHANNEL_WIDTH, MAX_TX_POWER, BSSID_C);
+		State stateC = TestUtils.createState(
+			1, DEFAULT_CHANNEL_WIDTH, MAX_TX_POWER, BSSID_C
+		);
```
elludraon, line 93: Is there a reason why `stateC` is created differently from the others?

elludraon (approved):
Looks good, please address minor comments before pushing

src/test/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPCTest.java
RockyMandayam2: Good catch. Actually, this needs some more cleaning up, which I'll do after vacation :P For now, I'm putting stateC back to how it was.

PR #57, opened 2022-08-30
Implement a POST/system command required by OpenWIFI SDK

zhiqiand:
This PR implemented a POST command required by OpenWifi SDK. 
- In `ApiServer.java`, add a new EndPoint `SetSystemEndpoint` to implement POST.
 -- requestBody: Parse this as a JSONObject as  the value of Schema.implementation 
--  responses: Content of `Schema` is Object.class and leave a TODO comment in the code. Simply return "[]" for each valid command.

- Add unit test in `ApiServerTest.java` to test the changes.
- openapi.yaml is auto-generated

Test plan:

1. valid command: `{"command": "getloglevelnames"}`

Screen Shot 2022-08-30 at 2 23 39 PM

2. invalid command: `{"command": "xx"}`
Screen Shot 2022-08-30 at 2 23 59 PM
3. invalid command: : `{"xx": "xx"}`
Screen Shot 2022-08-30 at 2 24 15 PM

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -325,6 +325,29 @@ paths:
           description: Success
         "400":
           description: Bad Request
+  /api/v1/setSystem:
+    post:
+      tags:
+      - SDK
+      summary: System commands details
+      description: Perform some system wide commands.
+      operationId: systemCommand
+      requestBody:
+        description: Command details
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/JSONObject'
+        required: true
+      responses:
+        "200":
+          description: Successful command execution
+          content:
+            application/json:
+              schema:
+                type: object
+        "400":
+          description: Bad Request
   /api/v1/setTopology:
     post:
       tags:
@@ -527,6 +550,11 @@ components:
             additionalProperties:
               type: integer
               format: int32
+    JSONObject:
+      type: object
+      properties:
+        empty:
+          type: boolean
     Certificate:
       type: object
       properties:
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -325,19 +325,43 @@ paths:
           description: Success
         "400":
           description: Bad Request
-  /api/v1/setSystem:
+  /api/v1/system:
+    get:
+      tags:
+      - SDK
+      summary: Get system info
+      description: Returns the system info from the running service.
+      operationId: system
+      parameters:
+      - name: command
+        in: query
+        description: Get a value
+        required: true
+        schema:
+          type: string
+          enum:
+          - info
+      responses:
+        "200":
+          description: Success
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/SystemInfoResults'
+        "400":
+          description: Bad Request
     post:
       tags:
       - SDK
-      summary: System commands details
-      description: Perform some system wide commands.
-      operationId: systemCommand
+      summary: Run system commands
+      description: Perform some system-wide commands.
+      operationId: setSystem
       requestBody:
         description: Command details
         content:
           application/json:
             schema:
-              $ref: '#/components/schemas/JSONObject'
+              type: object
         required: true
       responses:
         "200":
@@ -365,31 +389,6 @@ paths:
       responses:
         "200":
           description: Success
-        "400":
-          description: Bad Request
-  /api/v1/system:
-    get:
-      tags:
-      - SDK
-      summary: Get system info
-      description: Returns the system info from the running service.
-      operationId: system
-      parameters:
-      - name: command
-        in: query
-        description: Get a value
-        required: true
-        schema:
-          type: string
-          enum:
-          - info
-      responses:
-        "200":
-          description: Success
-          content:
-            application/json:
-              schema:
-                $ref: '#/components/schemas/SystemInfoResults'
         "400":
           description: Bad Request
 components:
@@ -550,11 +549,6 @@ components:
             additionalProperties:
               type: integer
               format: int32
-    JSONObject:
-      type: object
-      properties:
-        empty:
-          type: boolean
     Certificate:
       type: object
       properties:
--- a/src/main/java/com/facebook/openwifirrm/modules/ApiServer.java
+++ b/src/main/java/com/facebook/openwifirrm/modules/ApiServer.java
@@ -30,6 +30,7 @@ import javax.ws.rs.Path;
 import javax.ws.rs.Produces;
 import javax.ws.rs.core.MediaType;
 
+import org.json.JSONObject;
 import org.reflections.Reflections;
 import org.reflections.util.ConfigurationBuilder;
 import org.slf4j.Logger;
@@ -212,6 +213,7 @@ public class ApiServer implements Runnable {
 		Spark.after(this::afterFilter);
 		Spark.options("/*", this::options);
 		Spark.get("/api/v1/system", new SystemEndpoint());
+		Spark.post("/api/v1/setSystem", new SetSystemEndpoint());
 		Spark.get("/api/v1/provider", new ProviderEndpoint());
 		Spark.get("/api/v1/algorithms", new AlgorithmsEndpoint());
 		Spark.put("/api/v1/runRRM", new RunRRMEndpoint());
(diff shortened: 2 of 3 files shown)


openapi.yaml
```diff
@@ -527,6 +550,11 @@ components:
             additionalProperties:
               type: integer
               format: int32
+    JSONObject:
```
RockyMandayam2, line 553: Can we name this something relevant like `SystemDetails`? Even if nothing is implemented yet?

openapi.yaml
```diff
@@ -325,6 +325,29 @@ paths:
           description: Success
         "400":
           description: Bad Request
+  /api/v1/setSystem:
+    post:
+      tags:
+      - SDK
+      summary: System commands details
```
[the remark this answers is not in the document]
zhiqiand: `openapi.yaml` is automatically generated. I followed the summary/description with the[ ucentralgw repo ](https://github.com/Telecominfraproject/wlan-cloud-ucentralgw/blob/f014960b3a9e5bedb6dbc57fde788d299bc5cf80/openapi/owgw.yaml#L2619-L2649). I could also reverted it to match our own coding style.

openapi.yaml
zhiqiand: I simply implemented it as [T129272616](https://www.internalfb.com/intern/tasks/?t=129272616) says. But I am ok with both way. Maybe I could do it separately. what's your opinion @elludraon ?

openapi.yaml
```diff
@@ -527,6 +550,11 @@ components:
             additionalProperties:
               type: integer
               format: int32
+    JSONObject:
+      type: object
+      properties:
```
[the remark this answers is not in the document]
zhiqiand: As mentioned before, this file is auto-generated. It might be related to some properties owned by JSONObject. My understanding is that "command" is a string field under requestBody. (Not sure but correct me if I'm wrong)
Screen Shot 2022-08-30 at 11 40 24 AM

openapi.yaml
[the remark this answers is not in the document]
RockyMandayam2: Oh, in that case, I think it is fine as is

openapi.yaml
RockyMandayam2: Oh, I see, it's fine as is then.

src/main/java/com/facebook/openwifirrm/modules/ApiServer.java
```diff
@@ -520,6 +522,51 @@ public String handle(
...
 	}
+	
+	@Path("/api/v1/setSystem")
+	public class SetSystemEndpoint implements Route {
+		@POST
+		@Produces({ MediaType.APPLICATION_JSON })
+		@Operation(
+			summary = "System commands details", 
+			description = "Perform some system wide commands.", 
+			operationId = "systemCommand",
```
elludraon, line 533:
```suggestion
			summary = "Run system commands", 
			description = "Perform some system-wide commands.", 
			operationId = "setSystem", 
```

src/main/java/com/facebook/openwifirrm/modules/ApiServer.java
```diff
@@ -520,6 +522,51 @@ public String handle(
...
+			summary = "System commands details", 
+			description = "Perform some system wide commands.", 
+			operationId = "systemCommand", 
+			tags = {"SDK" }, 
+			requestBody = @RequestBody(
+				description = "Command details", 
+				content = {
+					@Content(
+						mediaType = "application/json", 
+						schema = @Schema(implementation = JSONObject.class))
```
elludraon, line 540: Use `Object.class` here too, it is already implied JSON due to "mediaType"

src/test/java/com/facebook/openwifirrm/modules/ApiServerTest.java
```diff
@@ -559,4 +559,25 @@ void test_runRRM() throws Exception {
 		assertEquals(400, Unirest.put(url + "?mode=test123&venue=" + zone).asString().getStatus());
 		assertEquals(400, Unirest.put(url + "?venue=asdf&algorithm=" + algorithms.get(0)).asString().getStatus());
 	}
-}
+
+
+	@Test
+	@Order(2004)
+	void test_setSystem() throws Exception {
```
elludraon, line 566: nit: could you move this to just below `test_system()` and renumber the tests under it?

src/test/java/com/facebook/openwifirrm/modules/ApiServerTest.java
```diff
@@ -559,4 +559,25 @@ void test_runRRM() throws Exception {
...
+		String url = endpoint("/api/v1/setSystem");
```
elludraon, line 567: `/api/v1/system`

src/main/java/com/facebook/openwifirrm/modules/ApiServer.java
```diff
@@ -520,6 +522,51 @@ public String handle(
...
+				@Parameter(hidden = true) Request request,
+				@Parameter(hidden = true) Response response
+		) {
+			try {
+				JSONObject obj = gson.fromJson(request.body(), JSONObject.class);
+			} catch (Exception e) {
+				response.status(400);
+				return e.getMessage();
+			}
+			return "";
```
elludraon, line 567: I expected a bit more in this function body, specifically:
- Parse the "command" field from the request JSON object and match it against allowed values (this was defined in the upstream yaml file): `setloglevel, reload, getloglevels, getloglevelnames, getsubsystemnames`
- Return empty array on success (instead of empty string)

```java
switch (command) {
case "setloglevel":
case "<... fill in the rest>":
  return "[]";
default:
  response.status(400);
  return "Invalid command";
}
```

src/main/java/com/facebook/openwifirrm/modules/ApiServer.java
```diff
@@ -520,6 +522,51 @@ public String handle(
...
+				@ApiResponse(responseCode = "400", description = "Bad Request")
+			}
+		)
+		@Override
+		public String handle(
+				@Parameter(hidden = true) Request request,
+				@Parameter(hidden = true) Response response
+		) {
+			try {
+				JSONObject obj = gson.fromJson(request.body(), JSONObject.class);
```
elludraon, line 562: Don't mix `JSONObject` and `Gson` - they are two different libraries (the gson library's version is called `JsonObject`, which may have been confusing). You can choose to use either library here.

```java
// gson
JsonObject obj = gson.fromJson(request.body(), JsonObject.class);

// org.json
JSONObject obj = new JSONObject(request.body());
```

src/test/java/com/facebook/openwifirrm/modules/ApiServerTest.java
```diff
@@ -559,4 +559,25 @@ void test_runRRM() throws Exception {
...
+
+		// Create JSONObject
+		JSONObject json_obj = new JSONObject();
+		json_obj.put("aa", 'a');
+		json_obj.put("bb", 'b');
```
elludraon, line 572: I would just write this manually as a string literal since it's simple, like:
```java
"{\"command\": \"reload\"}"
```

src/test/java/com/facebook/openwifirrm/modules/ApiServerTest.java
```diff
@@ -559,4 +559,25 @@ void test_runRRM() throws Exception {
...
+		
+		HttpResponse<String> resp = Unirest
+			.post(url)
+			.body(gson.toJson(json_obj))
```
elludraon, line 576: See comments in `ApiServer` above - you should be checking for a valid "command" field in order to return success. Add tests for missing/invalid command (in addition to invalid JSON) in the failure cases.

elludraon (changes requested):

src/main/java/com/facebook/openwifirrm/modules/ApiServer.java
```diff
@@ -520,6 +522,66 @@ public String handle(
...
+					case "setloglevel":
+						return "[]";
+					case "reload":
+						return "[]";
+					case "getloglevels":
+						return "[]";
+					case "getloglevelnames":
+						return "[]";
+					case "getsubsystemnames":
+						return "[]";
```
elludraon, line 574: Since we are handling them all the same way, you can write it like this:
```suggestion
					case "setloglevel":
					case "reload":
					case "getloglevels":
					case "getloglevelnames":
					case "getsubsystemnames":
						return "[]";
```

src/main/java/com/facebook/openwifirrm/modules/ApiServer.java
```diff
@@ -520,6 +522,66 @@ public String handle(
...
+				@ApiResponse(responseCode = "400", description = "Bad Request")
+			}
+		)
+		@Override
+		public String handle(
+				@Parameter(hidden = true) Request request,
+				@Parameter(hidden = true) Response response
+		) {
+			try {
+				JSONObject json_obj = new JSONObject(request.body());
```
elludraon, line 562: For variable naming in Java, usually we use camelCase, ex. `jsonObj`. Please also rename the variables in the test file.

src/test/java/com/facebook/openwifirrm/modules/ApiServerTest.java
```diff
@@ -507,9 +507,29 @@ void testCORS() throws Exception {
...
-		assertEquals(VersionProvider.get(), resp.getBody().getObject().getString("version"));
+		// Test on GET api
+		HttpResponse<JsonNode> get_resp = Unirest.get(endpoint("/api/v1/system?command=info")).asJson();
+		assertEquals(200, get_resp.getStatus());
+		assertEquals(VersionProvider.get(), get_resp.getBody().getObject().getString("version"));
+
+		// Test on POST api
+		String url = endpoint("/api/v1/system");
+		// Valid command
+		JSONObject json_obj = new JSONObject("{\"command\": \"reload\"}");
```
elludraon, line 518: To clarify, you *don't* need to create a `JSONObject` here (or the other ones below). Just pass it as a string literal in the function call.

elludraon (commented):
A few last style comments, then this is good to go!

elludraon (approved):

PR #58, opened 2022-08-30
Use current time for aggregation

RockyMandayam2:
This PR is based on a discussion I had with Po-Han and Haleema. Basically, we actually should be using the current time to determine obsoletion in wifiscan entry aggregation. However, if we do not have a recent entry (so there is nothing to aggregate over), we just return the most recent entry (without any aggregation). This PR updates the aggregation logic as well as the tests.

### DIFF
--- a/src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
+++ b/src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
@@ -334,7 +334,12 @@ public class ModelerUtils {
 				}
 				if (agg.getCount() > 0) {
 					aggregatedWifiScans.get(serialNumber).get(bssid).signal = (int) Math.round(agg.getAggregate());
+				} else {
+					aggregatedWifiScans
+						.computeIfAbsent(serialNumber, k -> new HashMap<>())
+						.put(bssid, mostRecentEntry);
 				}
+
 			}
 		}
 		return aggregatedWifiScans;
--- a/src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
+++ b/src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
@@ -237,28 +237,34 @@ public class ModelerUtils {
 	}
 
 	/**
-	 * For each AP, for each other AP that sent a wifiscan entry to that AP, this
-	 * method calculates an aggregate wifiscan entry with an aggregated RSSI.
+	 * For each AP, for each other AP that sent a wifiscan entry to that AP,
+	 * this method calculates an aggregate wifiscan entry with an aggregated
+	 * RSSI. If no non-obsolete entry exists, the latest wifiscan entry is used
+	 * instead.
 	 *
-	 * @param dataModel          the data model which includes the latest wifiscan
-	 *                           entries
-	 * @param obsoletionPeriodMs for each (scanning AP, responding AP) tuple, the
-	 *                           maximum amount of time (in milliseconds) it is
-	 *                           worth aggregating over, starting from the most
-	 *                           recent scan entry for that tuple, and working
-	 *                           backwards in time. An entry exactly
-	 *                           {@code obsoletionPeriodMs} ms earlier than the most
-	 *                           recent entry is considered non-obsolete (i.e., the
-	 *                           "non-obsolete" window is inclusive). Must be
-	 *                           non-negative.
+	 * @param dataModel          the data model which includes the latest
+	 *                           wifiscan entries
+	 * @param obsoletionPeriodMs for each (scanning AP, responding AP) tuple,
+	 *                           the maximum amount of time (in milliseconds) it
+	 *                           is worth aggregating over, starting from the
+	 *                           most recent scan entry for that tuple, and
+	 *                           working backwards in time. An entry exactly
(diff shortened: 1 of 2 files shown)


src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -236,32 +237,59 @@ private static boolean matchesForAggregation(WifiScanEntry entry1, WifiScanEntry
...
-	 *                           maximum amount of time (in milliseconds) it is
-	 *                           worth aggregating over, starting from the most
-	 *                           recent scan entry for that tuple, and working
-	 *                           backwards in time. An entry exactly
-	 *                           {@code obsoletionPeriodMs} ms earlier than the most
-	 *                           recent entry is considered non-obsolete (i.e., the
-	 *                           "non-obsolete" window is inclusive). Must be
-	 *                           non-negative.
+	 * @param dataModel          the data model which includes the latest
+	 *                           wifiscan entries
```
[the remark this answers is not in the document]
RockyMandayam2: I accidentally used my eclipse formatter.  It's ok, since Jeffrey is going to format the repo soon anyways, I won't worry about format too much in this PR.

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -236,32 +237,59 @@ private static boolean matchesForAggregation(WifiScanEntry entry1, WifiScanEntry
...
+	public static Map<String, Map<String, WifiScanEntry>> getAggregatedWifiScans(
+		Modeler.DataModel dataModel,
+		long obsoletionPeriodMs,
+		Aggregator<Double> agg
+	) {
+		return getAggregatedWifiScans(
+			dataModel,
+			obsoletionPeriodMs,
+			new MeanAggregator(),
+			System.currentTimeMillis()
```
zhiqiand, line 278: I might have some misunderstanding on the current time. Why we use current system time here? For aggregation, my understanding is that we use the time when the AP received signals. If we use current system time, how can we minimize the measurement error?

RockyMandayam2: The current time is used to measure against to figure out how obsolete an entry is. E.g., if it is 3pm and we received an entry at 1pm, that is 2 hours old (and according to the current 15 min obsoletion period, that makes it obsolete).

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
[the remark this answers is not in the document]
elludraon: The autoformatter is not going to touch comments, so format them yourself if needed (I don't really care as long as it's readable)

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -236,32 +237,59 @@ private static boolean matchesForAggregation(WifiScanEntry entry1, WifiScanEntry
...
+		);
+	}
+
+	/**
+	 * @see #getAggregatedWifiScans(com.facebook.openwifirrm.modules.Modeler.DataModel, long, Aggregator)
+	 * <p>
+	 * These two methods were separated to make testing easier, since it relies on current time.
```
elludraon, line 285: - `@see` should be at the end, and you need a description first, ex. "Compute aggregated wifiscans using a given reference time"
- The note here does not belong in Javadoc, move it to inline comment

src/test/java/com/facebook/openwifirrm/modules/ModelerUtilsTest.java
```diff
@@ -99,11 +99,20 @@ void testPreDot11nAggregatedWifiScanEntry() {
 		final String bssidC = "cc:cc:cc:cc:cc:cc";
 		DataModel dataModel = new DataModel();
 
+		long currentUnixTimeMs = TestUtils.DEFAULT_WIFISCANENTRY_TIME.toEpochMilli();
```
elludraon, line 102: rename this `refTimeMs` because it is explicitly not the current time

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -236,32 +237,59 @@ private static boolean matchesForAggregation(WifiScanEntry entry1, WifiScanEntry
...
+	 */
+	public static Map<String, Map<String, WifiScanEntry>> getAggregatedWifiScans(
+		Modeler.DataModel dataModel,
+		long obsoletionPeriodMs,
+		Aggregator<Double> agg,
+		long currentUnixTimeMs
```
elludraon, line 291: Suggest renaming `currentUnixTimeMs` to `refTimeMs` since you are intending to not always pass in current time

elludraon (changes requested):
Few minor comments, requesting changes since I think you added invalid javadoc.

RockyMandayam2:
Comments addressed, and I fixed javadoc issues (`mvn javadoc:javadoc` passes now, forgot to run it earlier).

elludraon (approved):

PR #60, opened 2022-08-31
Autoformat codebase using Spotless

elludraon:
Autoformat codebase using `mvn spotless:apply`, see #49 for details.

`mvn test` and `mvn javadoc:javadoc` still pass.


src/main/java/com/facebook/openwifirrm/DeviceDataManager.java
```diff
@@ -31,7 +31,8 @@
  * Device topology and config manager.
  */
 public class DeviceDataManager {
-	private static final Logger logger = LoggerFactory.getLogger(DeviceDataManager.class);
+	private static final Logger logger =
```
RockyMandayam2, line 34: Noted, this is how we do things going forward

RockyMandayam2 (approved):

PR #61, opened 2022-08-31
Pass down RRM algorithm parameters

jwoos:
Args were being parsed out but not actually passed into the constructors. Create and call a factory method to call the proper constructors based on what's present in the generic args map for all the algorithms available. Also updated ALGORITHMS.md with the available args for each algorithm.

### DIFF
--- a/ALGORITHMS.md
+++ b/ALGORITHMS.md
@@ -11,9 +11,8 @@ selected channel. This is only for testing and re-initialization.
 
 Parameters:
 * `mode`: "random"
-* `args`:
-  * `setDifferentChannelPerAp`: If true, will set a different random channel per AP. If false, it will set the same random channel for all APs.
-    * values: true | false (default: false)
+* `setDifferentChannelPerAp`: If true, will set a different random channel per AP. If false, it will set the same random channel for all APs.
+    * values: `true`, `false` (default: `false`)
 
 ### `LeastUsedChannelOptimizer`
 This algorithm assigns the channel of the OWF APs based on the following logic:
@@ -29,7 +28,6 @@ This algorithm assigns the channel of the OWF APs based on the following logic:
 
 Parameters:
 * `mode`: "least_used"
-* `args:`
 
 ### `UnmanagedApAwareChannelOptimizer`
 Building on the least used channel assignment algorithm, this algorithm can
@@ -52,7 +50,6 @@ are based on the following logic:
 
 Parameters:
 * `mode`: "unmanaged_aware"
-* `args:`
 
 ## Transmit Power Control
 `TPC` and its subclasses implement various transmit power control algorithms,
@@ -64,9 +61,8 @@ the value. This is only for testing and re-initialization.
 
 Parameters:
 * `mode`: "random"
-* `args`:
-  * `setDifferentTxPowerPerAp`: If true, will set a different random tx power per AP. If false, it will set the same random tx power for all APs.
-    * values: true | false (default: false)
+* `setDifferentTxPowerPerAp`: If true, will set a different random tx power per AP. If false, it will set the same random tx power for all APs.
+    * values: `true`, `false` (default: `false`)
 
 ### `MeasurementBasedApClientTPC`
 This algorithm tries to assign the Tx power of the OWF APs based on the
@@ -80,8 +76,7 @@ each AP):
 
 Parameters:
 * `mode`: "measure_ap_client"
-* `args`:
-  * `targetMcs`: The target MCS index
+* `targetMcs`: The target MCS index
     * values: 0-9 (default: 8)
 
 ### `MeasurementBasedApApTPC`
@@ -98,8 +93,7 @@ levels of these APs will be determined by the following steps:
 
 Parameters:
 * `mode`: "measure_ap_ap"
-* `args`:
-  * `coverageThreshold`: Coverage threshold between APs in dBm
+* `coverageThreshold`: Coverage threshold between APs in dBm
     * values:  int < 30 (default: -70)
-  * `nthSmallestRssi`: the nth smallest RSSI that is used for tx power calculation
+* `nthSmallestRssi`: the nth smallest RSSI that is used for tx power calculation
     * values: int >= 0 (default: 0)
--- a/ALGORITHMS.md
+++ b/ALGORITHMS.md
@@ -11,6 +11,9 @@ selected channel. This is only for testing and re-initialization.
 
 Parameters:
 * `mode`: "random"
+* `args`:
+  * `setDifferentChannelPerAp`: If true, will set a different random channel per AP. If false, it will set the same random channel for all APs.
+    * values: true | false (default: false)
 
 ### `LeastUsedChannelOptimizer`
 This algorithm assigns the channel of the OWF APs based on the following logic:
@@ -26,6 +29,7 @@ This algorithm assigns the channel of the OWF APs based on the following logic:
 
 Parameters:
 * `mode`: "least_used"
+* `args:`
 
 ### `UnmanagedApAwareChannelOptimizer`
 Building on the least used channel assignment algorithm, this algorithm can
@@ -48,6 +52,7 @@ are based on the following logic:
 
 Parameters:
 * `mode`: "unmanaged_aware"
+* `args:`
 
 ## Transmit Power Control
 `TPC` and its subclasses implement various transmit power control algorithms,
@@ -59,6 +64,9 @@ the value. This is only for testing and re-initialization.
 
 Parameters:
 * `mode`: "random"
+* `args`:
+  * `setDifferentTxPowerPerAp`: If true, will set a different random tx power per AP. If false, it will set the same random tx power for all APs.
+    * values: true | false (default: false)
 
 ### `MeasurementBasedApClientTPC`
 This algorithm tries to assign the Tx power of the OWF APs based on the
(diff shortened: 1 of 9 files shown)


RockyMandayam2:
Is there any way to test this?

ALGORITHMS.md
```diff
@@ -26,6 +29,7 @@ This algorithm assigns the channel of the OWF APs based on the following logic:
 
 Parameters:
 * `mode`: "least_used"
+* `args:`
```
elludraon, line 32: omit if empty

ALGORITHMS.md
```diff
@@ -11,6 +11,9 @@ selected channel. This is only for testing and re-initialization.
 
 Parameters:
 * `mode`: "random"
+* `args`:
+  * `setDifferentChannelPerAp`: If true, will set a different random channel per AP. If false, it will set the same random channel for all APs.
+    * values: true | false (default: false)
```
elludraon, line 16: - "args" shouldn't be a separate section, it should all go under "Parameters" ("mode" is passed the same way)
- You need to indent list items with 4 spaces (not 2) for better Markdown renderer support (from what I understand)
- see suggested styling for values (commas, code font)

```suggestion
* `setDifferentChannelPerAp`: If true, will set a different random channel per AP. If false, it will set the same random channel for all APs.
    * Values: `true`, `false` (default: `false`)
```

src/main/java/com/facebook/openwifirrm/optimizers/channel/RandomChannelInitializer.java
```diff
@@ -38,11 +38,38 @@ public class RandomChannelInitializer extends ChannelOptimizer {
...
+	/** Factory method to parse generic args map into the proper constructor */
+	public static RandomChannelInitializer makeWithArgs(
+		DataModel model,
+		String zone,
+		DeviceDataManager deviceDataManager,
+		Map<String, String> args
+	) {
+		if (args.containsKey(ARG_SET_DIFFERENT_CHANNEL_PER_AP)) {
+			boolean setDifferentChannelPerAp =
+				args.get(ARG_SET_DIFFERENT_CHANNEL_PER_AP) == "true";
```
elludraon, line 62: - Use String `.equals()`, this is Java 😛 
- I mildly prefer if you don't define class constants for the arg names, and just write them inline in this method to keep it all in one place. I don't suspect they'll be needed elsewhere.
- Avoid double lookups in maps.
- To make it easier to identify default values, I suggest avoiding the if-else blocks to call different constructors for their defaults, and instead define the defaults explicitly.

Suggestion:
```java
// Declare args with defaults at the top
boolean setDifferentChannelPerAp = false;

// Parse and validate args
String arg;
if ((arg = args.get("setDifferentChannelPerAp")) != null) {
  if (arg.equals("true")) {
    setDifferentChannelPerAp = true;
  }
}
if ((arg = args.get("something_else")) != null) {
  /* ... */
}

// Use one constructor only
return new <...>;
```

src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPC.java
```diff
@@ -58,12 +58,51 @@ public class MeasurementBasedApApTPC extends TPC {
...
+		DataModel model,
+		String zone,
+		DeviceDataManager deviceDataManager,
+		Map<String, String> args
+	) {
+
+		if (
+			args.containsKey(ARG_COVERAGE_THRESHOLD) &&
+				args.containsKey(ARG_NTH_SMALLEST_RSSI)
+		) {
```
elludraon, line 88: We should accept any number of args, not all-or-nothing.

elludraon (changes requested):
Needs a few changes, but I do like the design choice.

src/main/java/com/facebook/openwifirrm/optimizers/channel/RandomChannelInitializer.java
elludraon: Also note you can use `Boolean.parseBoolean()`

src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPC.java
```diff
@@ -58,12 +58,51 @@ public class MeasurementBasedApApTPC extends TPC {
...
+			int coverageThreshold =
+				Integer.parseInt(args.get(ARG_COVERAGE_THRESHOLD));
```
elludraon, line 90: I would wrap all of this in a try-catch block for `NumberFormatException` so you don't crash on bad args. Log an error and keep the default value if that happens.

src/main/java/com/facebook/openwifirrm/optimizers/channel/RandomChannelInitializer.java
```diff
@@ -38,33 +38,63 @@ public class RandomChannelInitializer extends ChannelOptimizer {
...
+	public static RandomChannelInitializer makeWithArgs(
+		DataModel model,
+		String zone,
+		DeviceDataManager deviceDataManager,
+		Map<String, String> args
+	) {
+		boolean setDifferentChannelPerAp = DEFAULT_SET_DIFFERENT_CHANNEL_PER_AP;
+
+		String arg;
+		if ((arg = args.get("rrm-algorithm-params")) != null) {
```
elludraon, line 60: copy/paste typo

src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPC.java
```diff
@@ -64,6 +64,48 @@ public class MeasurementBasedApApTPC extends TPC {
...
+		int coverageThreshold = DEFAULT_COVERAGE_THRESHOLD;
+		int nthSmallestRssi = DEFAULT_NTH_SMALLEST_RSSI;
+
+		String arg;
+		if ((arg = args.get("coverageThreshold")) != null) {
+			try {
+				coverageThreshold = Integer.parseInt(arg);
+			} catch (NumberFormatException e) {
+				logger.error(
+					"Invalid integer passed to parameter coverageThreshold, using default value: ",
```
elludraon, line 83: nit: for `logger.error` you don't need to end the string with `: `, it prints the exception in a readable way already

src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPC.java
```diff
@@ -64,6 +64,48 @@ public class MeasurementBasedApApTPC extends TPC {
...
+				);
+			}
+		}
+
+		if ((arg = args.get("nthSmallestRssi")) != null) {
+			try {
+				nthSmallestRssi = Integer.parseInt(arg);
+			} catch (NumberFormatException e) {
+				logger.error(
+					"Invalid integer passed to parameter nthSmallestRssi, using default value: ",
```
elludraon, line 94: same here

src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPC.java
```diff
@@ -64,6 +64,48 @@ public class MeasurementBasedApApTPC extends TPC {
...
+		DeviceDataManager deviceDataManager,
+		Map<String, String> args
+	) {
+		int coverageThreshold = DEFAULT_COVERAGE_THRESHOLD;
+		int nthSmallestRssi = DEFAULT_NTH_SMALLEST_RSSI;
+
+		String arg;
+		if ((arg = args.get("coverageThreshold")) != null) {
+			try {
+				coverageThreshold = Integer.parseInt(arg);
```
elludraon, line 80: also validate that this is in range? (set to temp variable first, do range checks, then assign `coverageThreshold`)

src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPC.java
```diff
@@ -64,6 +64,48 @@ public class MeasurementBasedApApTPC extends TPC {
...
+				logger.error(
+					"Invalid integer passed to parameter coverageThreshold, using default value: ",
+					e
+				);
+			}
+		}
+
+		if ((arg = args.get("nthSmallestRssi")) != null) {
+			try {
+				nthSmallestRssi = Integer.parseInt(arg);
```
elludraon, line 91: same here

src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApClientTPC.java
```diff
@@ -61,6 +61,42 @@ public class MeasurementBasedApClientTPC extends TPC {
...
+					e
+				);
+			}
+			if (targetMcs < 0) {
+				logger.error(
+					"Invalid value passed for targetMcs - must be greater than 0. Using default value."
+				);
+				targetMcs = DEFAULT_TARGET_MCS;
+			}
+		}
```
elludraon, line 90: nit: to avoid needing to reassign the default (which looks messy), I would set a temp variable first and do range checks on that. can also move this into the try-catch block if you want (to limit the temp var scope).

elludraon (changes requested):

elludraon (approved):

PR #62, opened 2022-08-31
test nonzero nthSmallestRssi

RockyMandayam2:


jwoos (approved):
Went through the algorithm document and made sure the test was correct - lgtm

PR #65, opened 2022-08-31
Use TPC user configs (allowedTxPowers, userTxPowers)

wuweicaifb:
The TPC algorithms did not use some user config fields from DeviceConfig (the device configuration):
- allowedTxPowers
- userTxPowers

This PR uses those field in TPC algorithms.

Here's the list of major updates and some questions:
1. For available channel lists, we have different default channel lists for different channel widths and bands, but for TPC we don't. I use all the tx powers in the range of [`MIN_TX_POWER`, `MAX_TX_POWER`] as default for now. Any idea to have more specific default? @pohanhf 
2. `RandomTxPowerIntializer`: since different AP and band may have different available tx powers now, `setDifferentTxPowerPerAp` is removed. If we want to keep this arg, we can pick random one from the common tx power choices.
3. `LocationBasedOptimalTPC`: leverage the new method when computing tx powers, but there's a little functionally different. `userTxPowers` is not considered before, but is used to update `txPowerChoices` in the new method.
4. `MeasurementBasedApClientTPC` and `MeasurementBasedApApTPC`: these two compute the tx power directly without any choices, so the new method is not used in these two optimizers. We can check if the computed tx power is valid, i.e. in the `txPowerChoices`, but that's not done in this PR.

### DIFF
--- a/src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApClientTPC.java
+++ b/src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApClientTPC.java
@@ -154,7 +154,8 @@ public class MeasurementBasedApClientTPC extends TPC {
 	private int computeTxPowerForRadio(
 		String serialNumber,
 		State state,
-		JsonObject radio
+		JsonObject radio,
+		List<Integer> txPowerChoices
 	) {
 		// Find current tx power and bandwidth
 		int currentTxPower =
@@ -166,6 +167,9 @@ public class MeasurementBasedApClientTPC extends TPC {
 				!radio.get("channel_width").isJsonNull()
 					? radio.get("channel_width").getAsInt()
 					: 20);
+		Collections.sort(txPowerChoices);
+		int minTxPower = txPowerChoices.get(0);
+		int maxTxPower = txPowerChoices.get(txPowerChoices.size() - 1);
 
 		// Find minimum client RSSI
 		List<Integer> clientRssiList = new ArrayList<>();
@@ -196,12 +200,12 @@ public class MeasurementBasedApClientTPC extends TPC {
 		// TODO: revisit this part to have a better logic
 		if (clientRssiList.isEmpty()) {
 			logger.info(
-				"Device {}: no clients, assigning default tx power {} (was {})",
+				"Device {}: no clients, assigning minimal possible tx power {} (was {})",
 				serialNumber,
-				DEFAULT_TX_POWER,
+				minTxPower,
 				currentTxPower
 			);
-			return DEFAULT_TX_POWER; // no clients
+			return minTxPower; // no clients
 		}
 		int clientRssi = Collections.min(clientRssiList);
 
@@ -226,12 +230,12 @@ public class MeasurementBasedApClientTPC extends TPC {
 				newTxPower
 			);
 
-			// If this exceeds max tx power, repeat for (MCS - 1)
-			if (newTxPower > MAX_TX_POWER) {
+			// If this exceeds max possible tx power, repeat for (MCS - 1)
+			if (newTxPower > maxTxPower) {
 				logger.info(
 					"Device {}: computed tx power > maximum {}, trying with mcs - 1",
 					serialNumber,
-					MAX_TX_POWER
+					maxTxPower
 				);
 				if (--mcs >= 0) {
 					continue;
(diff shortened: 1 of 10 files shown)


wuweicaifb:
Update `DEFAULT_TX_POWER_CHOICES` as [20, 30] as was discussed with @pohanhf offline.

elludraon (commented):
This looks okay to me, but maybe get a second review (@pohanhf?) for good measure.

src/main/java/com/facebook/openwifirrm/optimizers/tpc/RandomTxPowerInitializer.java
```diff
@@ -38,36 +39,21 @@ public class RandomTxPowerInitializer extends TPC {
 	/** The PRNG instance. */
 	private final Random rng;
 
-	/** Whether to set a different value per AP or use a single value for all APs */
-	private final boolean setDifferentTxPowerPerAp;
```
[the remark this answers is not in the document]
wuweicaifb: As was mentioned in the point 2 of the summary, if we want to keep it, the easiest way is to randomly pick one from the common choices if it's true. If no common choices, raise an error. @pohanhf any idea?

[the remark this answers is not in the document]
pohanhf: Thanks for the question. I think we can keep it and use your idea to randomly pick one from the common choices. If no common choices, then just report some error messages to the users.

pohanhf (commented):
It looks good to me. Thanks Wuwei for enabling this feature. Let's keep the setDifferentTxPowerPerAp arg so we have some flexibility for initialization.

src/main/java/com/facebook/openwifirrm/optimizers/tpc/RandomTxPowerInitializer.java
```diff
@@ -113,31 +115,50 @@ public RandomTxPowerInitializer(
...
+					txPowerChoices = updateTxPowerChoices(
+						band,
+						serialNumber,
+						txPowerChoices
+					);
+				}
+			}
+
+			defaultTxPower =
+				txPowerChoices.get(rng.nextInt(txPowerChoices.size()));
```
jwoos, line 135: doesn't this violate the choices available since the set of tx power choices won't necessarily be the same across APs due to reading user config? e.g. AP A returns [1, 2, 3] and B returns [4, 5, 6] (just an example). Depending on iteration order we'd be violating either A or B's requested power.

wuweicaifb: Yes, you're right. When `setDifferentTxPowerPerAp` is false, `txPowerChoices` is computed as the shared tx powers across different APs, and then `dwfaultTxPower` is picked from that choices. If there's no shared tx powers (like you case), the default is used.

wuweicaifb:
1. Leverage user configs in Measurement TPCs. Only pick TPC in the tx power choices. If the new TPC is not in the choice list, the nearest tx power is picked.

2. Change default tx power choices from [20, 30] to [0, 30]

src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApClientTPC.java
```diff
@@ -226,38 +230,55 @@ private int computeTxPowerForRadio(
...
 		} while (true);
 
+		int txPowerIndex = 0;
+		for (
+			txPowerIndex = 0;
+			txPowerIndex < txPowerChoices.size() &&
+				txPowerChoices.get(txPowerIndex) < newTxPower;
+			txPowerIndex++
+		) {
+
```
jwoos, line 272: did you mean to leave this block?

jwoos (approved):
left a small comment about an unused block, looks good otherwise.

src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApClientTPC.java
wuweicaifb: Yes. Just want `txPowerIndex` to be the first index where the element is larger than or equal to `newTxPower`. No further computation is needed in this block.

jwoos: makes sense, but I guess as a personal thing I would expect a for loop to be just giving iteration logic and then the check would come inside the block with a `break` or something

PR #66, opened 2022-09-01
Implement per-channel TPC

wuweicaifb:
Signed-off-by: Wuwei Cai <wuweicai@fb.com>

## Summary

Implement per-channel operation to all TPC optimizers. This PR also fixes the bug that TPC always run for both 2G and 5G bands for `LocationBasedOptimalTPC`, `MeasurementApApTPC` and `RandomTxPowerInitializer`.
1. New method `getApsPerChannel` in the base `TPC` class. 
2. For `LocationBasedOptimalTPC`, instead of running algorithm per band, only consider APs with the same channel.
3. For `MeasurementApApTPC` and `RandomTxPowerInitializer`, only runs TPC on bands with channels on each AP.

## Test Plan
1. Per-band TPC test ran as expected with real APs
2. Modified the unit tests and tested with the cases that not channels in both 2G and 5G bands are used for each device.

### DIFF
--- a/src/main/java/com/facebook/openwifirrm/optimizers/tpc/LocationBasedOptimalTPC.java
+++ b/src/main/java/com/facebook/openwifirrm/optimizers/tpc/LocationBasedOptimalTPC.java
@@ -15,6 +15,7 @@ import java.util.Map;
 import java.util.TreeMap;
 import java.util.stream.Collectors;
 
+import com.facebook.openwifirrm.ucentral.UCentralUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -22,7 +23,6 @@ import com.facebook.openwifirrm.DeviceConfig;
 import com.facebook.openwifirrm.DeviceDataManager;
 import com.facebook.openwifirrm.modules.Modeler.DataModel;
 import com.facebook.openwifirrm.modules.ModelerUtils;
-import com.facebook.openwifirrm.ucentral.UCentralConstants;
 import com.facebook.openwifirrm.ucentral.models.State;
 
 /**
@@ -153,17 +153,20 @@ public class LocationBasedOptimalTPC extends TPC {
 	/**
 	 * Calculate new tx powers for the given band.
 	 *
-	 * @param band       band
+	 * @param channel channel
+	 * @param serialNumbers The serial numbers of the APs with the channel
 	 * @param txPowerMap this map from serial number to band to new tx power
 	 *                   (dBm) must be passed in empty, and it is filled in by
 	 *                   this method with the new tx powers.
 	 */
-	private void buildTxPowerMapForBand(
-		String band,
+	private void buildTxPowerMapForChannel(
+		int channel,
+		List<String> serialNumbers,
 		Map<String, Map<String, Integer>> txPowerMap
 	) {
 		int numOfAPs = 0;
 		int boundary = 100;
+		String band = UCentralUtils.getBandFromChannel(channel);
 		Map<String, Integer> validAPs = new TreeMap<>();
 		List<Double> apLocX = new ArrayList<>();
 		List<Double> apLocY = new ArrayList<>();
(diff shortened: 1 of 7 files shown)


src/main/java/com/facebook/openwifirrm/optimizers/tpc/TPC.java
```diff
@@ -178,4 +183,39 @@ public void applyConfig(
 		// Trigger config update now
 		configManager.wakeUp();
 	}
+
+	/**
+	 * Get AP serial numbers per channel.
+	 *
+	 * @return the map of channel to the list of serial numbers
+	 */
+	protected Map<Integer, List<String>> getApsPerChannel() {
```
jwoos, line 192: same comment as above Set vs LIst

src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPC.java
```diff
@@ -326,21 +325,24 @@ protected static int computeTxPower(
 	/**
 	 * Calculate new tx powers for the given band.
```
jwoos, line 326: can we update this since now it's taking in a channel? I know that channel maps to band but would be good to be explicit in the doc

src/main/java/com/facebook/openwifirrm/optimizers/tpc/LocationBasedOptimalTPC.java
```diff
@@ -15,14 +15,14 @@
 import java.util.TreeMap;
 import java.util.stream.Collectors;
 
+import com.facebook.openwifirrm.ucentral.UCentralUtils;
```
jwoos, line 18: did you mean to move this? Also it seems like we separate our imports vs third-party

jwoos (approved):
Logic seems to make sense to me, just some small comments

src/main/java/com/facebook/openwifirrm/optimizers/tpc/LocationBasedOptimalTPC.java
wuweicaifb: No. No idea why it was moved. Maybe I did it by mistake.

src/main/java/com/facebook/openwifirrm/optimizers/tpc/LocationBasedOptimalTPC.java
```diff
@@ -153,27 +153,29 @@ public static List<Integer> runLocationBasedOptimalTPC(
...
+	 * @param serialNumbers The serial numbers of the APs with the channel
 	 * @param txPowerMap this map from serial number to band to new tx power
 	 *                   (dBm) must be passed in empty, and it is filled in by
 	 *                   this method with the new tx powers.
 	 */
-	private void buildTxPowerMapForBand(
-		String band,
+	private void buildTxPowerMapForChannel(
+		int channel,
+		List<String> serialNumbers,
```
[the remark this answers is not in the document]
wuweicaifb: The list of serialNumbers are computed by `TPC.getApsPerChannel`, and I believe there would be no repetitions if our input radio state is good. Set is better in terms repetition as you said, but since we only use this collections for looping, a list is more efficient in both time and space in this use case.

src/main/java/com/facebook/openwifirrm/optimizers/tpc/LocationBasedOptimalTPC.java
wuweicaifb: Seems like my Intellij IDEA automatically put the new import here. Checking my config now.

src/main/java/com/facebook/openwifirrm/optimizers/tpc/LocationBasedOptimalTPC.java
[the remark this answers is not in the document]
jwoos: fair point! Yeah it was a minor comment more than anything

jwoos (approved):
LGTM!

PR #67, opened 2022-09-07
[WIFI-10750] use new API and schema to get RRM configs

jwoos:
Use the new API available in prov, which provides resolved RRM config per AP. This required some ... hacks.

- Create a new method that uses per device scheduling
- Upon API calls to RRM that modify the topology or settings, sync with prov to get the correct per AP setting

The above will be removed once the correct API is provided (venue level RRM settings as well as per device RRM settings that is unresolved)

### DIFF
--- a/src/main/java/com/facebook/openwifirrm/DeviceConfig.java
+++ b/src/main/java/com/facebook/openwifirrm/DeviceConfig.java
@@ -30,9 +30,6 @@ public class DeviceConfig {
 	/** The RRM schedule (if enabled), only applied at the zone or network layer */
 	public RRMSchedule schedule;
 
-	/** earameters for the algorithms, mapping algorithm name to string of a list of parameters */
-	public Map<String, String> algorithmsParameters;
-
 	/** Whether pushing device config is enabled */
 	public Boolean enableConfig;
 
@@ -84,7 +81,6 @@ public class DeviceConfig {
 
 		config.enableRRM = true;
 		config.schedule = null;
-		config.algorithmsParameters = null;
 		config.enableConfig = true;
 		config.enableWifiScan = true;
 		config.boundary = null;
--- a/src/main/java/com/facebook/openwifirrm/DeviceConfig.java
+++ b/src/main/java/com/facebook/openwifirrm/DeviceConfig.java
@@ -30,6 +30,9 @@ public class DeviceConfig {
 	/** The RRM schedule (if enabled), only applied at the zone or network layer */
 	public RRMSchedule schedule;
 
+	/** earameters for the algorithms, mapping algorithm name to string of a list of parameters */
+	public Map<String, String> algorithmsParameters;
+
 	/** Whether pushing device config is enabled */
 	public Boolean enableConfig;
 
@@ -81,6 +84,7 @@ public class DeviceConfig {
 
 		config.enableRRM = true;
 		config.schedule = null;
+		config.algorithmsParameters = null;
 		config.enableConfig = true;
 		config.enableWifiScan = true;
 		config.boundary = null;
--- a/src/main/java/com/facebook/openwifirrm/DeviceDataManager.java
+++ b/src/main/java/com/facebook/openwifirrm/DeviceDataManager.java
@@ -550,10 +550,10 @@ public class DeviceDataManager {
 		saveDeviceLayeredConfig();
 	}
 
-	/** Device AP config layer update interface. */
+	/** Device zone config layer update interface. */
 	public interface ZoneConfigFunction {
-		/** Update the AP config layer. */
-		void update(Map<String, DeviceConfig> apConfig);
+		/** Update the zone config layer. */
+		void update(Map<String, DeviceConfig> zoneConfig);
 	}
 
 	/**
--- a/src/main/java/com/facebook/openwifirrm/DeviceDataManager.java
+++ b/src/main/java/com/facebook/openwifirrm/DeviceDataManager.java
@@ -455,14 +455,14 @@ public class DeviceDataManager {
 	}
 
 	/**
-	 * Return the serail number of devices in a given zone.
+	 * Return config (with all config layers applied) for all devices in a given
+	 * zone, or null if not present in the topology.
 	 *
 	 * @param zone the zone that will be looked up for devi es
 	 *
-	 * @return set of serial numbers
+	 * @return map of serial number to computed config
 	 */
-	private Set<String> getAllDevicesInZone(String zone) {
-		// Get all devices in zone
+	public Map<String, DeviceConfig> getAllDeviceConfigs(String zone) {
 		if (zone == null || zone.isEmpty()) {
 			return null;
 		}
@@ -477,20 +477,6 @@ public class DeviceDataManager {
 		if (devicesInZone == null) {
 			return null;
 		}
-
-		return devicesInZone;
-	}
-
-	/**
-	 * Return config (with all config layers applied) for all devices in a given
-	 * zone, or null if not present in the topology.
-	 *
-	 * @param zone the zone that will be looked up for devi es
-	 *
-	 * @return map of serial number to computed config
-	 */
-	public Map<String, DeviceConfig> getAllDeviceConfigs(String zone) {
-		Set<String> devicesInZone = getAllDevicesInZone(zone);
 
 		// Compute config for all devices
 		Map<String, DeviceConfig> configMap = new HashMap<>();
@@ -565,13 +551,37 @@ public class DeviceDataManager {
 	}
 
 	/** Device AP config layer update interface. */
-	public interface DeviceApConfigFunction {
+	public interface LayerConfigFunction {
 		/** Update the AP config layer. */
 		void update(Map<String, DeviceConfig> apConfig);
 	}
 
-	/** Apply updates to the device AP config layer (under a write lock). */
-	public void updateDeviceApConfig(DeviceApConfigFunction fn) {
+	/**
+	 * Apply updates to the zone layer (under a write lock)
+	 *
+	 * @param fn The function to apply for the configs. The key of the argument is
+	 *        the zone name.
+	 */
+	public void updateZoneConfig(LayerConfigFunction fn) {
+		Lock l = deviceLayeredConfigLock.writeLock();
+		l.lock();
+		try {
+			fn.update(deviceLayeredConfig.zoneConfig);
+			sanitizeDeviceLayeredConfig(deviceLayeredConfig);
+		} finally {
+			l.unlock();
+		}
+		cachedDeviceConfigs.clear();
+		saveDeviceLayeredConfig();
+	}
+
+	/**
+	 * Apply updates to the device AP config layer (under a write lock).
+	 *
+	 * @param fn The function to apply for the configs. The key of the argument is
+	 *        the serial number of the device.
+	 */
+	public void updateDeviceApConfig(LayerConfigFunction fn) {
 		Lock l = deviceLayeredConfigLock.writeLock();
 		l.lock();
 		try {
--- a/src/main/java/com/facebook/openwifirrm/DeviceDataManager.java
+++ b/src/main/java/com/facebook/openwifirrm/DeviceDataManager.java
@@ -454,13 +454,7 @@ public class DeviceDataManager {
 		);
 	}
 
-	/**
-	 * Return config (with all config layers applied) for all devices in a given
-	 * zone, or null if not present in the topology.
-	 *
-	 * @return map of serial number to computed config
-	 */
-	public Map<String, DeviceConfig> getAllDeviceConfigs(String zone) {
+	private Set<String> getAllDevicesInZone(String zone) {
 		// Get all devices in zone
 		if (zone == null || zone.isEmpty()) {
 			return null;
@@ -476,6 +470,18 @@ public class DeviceDataManager {
 		if (devicesInZone == null) {
 			return null;
 		}
+
+		return devicesInZone;
+	}
+
+	/**
+	 * Return config (with all config layers applied) for all devices in a given
+	 * zone, or null if not present in the topology.
+	 *
+	 * @return map of serial number to computed config
+	 */
+	public Map<String, DeviceConfig> getAllDeviceConfigs(String zone) {
+		Set<String> devicesInZone = getAllDevicesInZone(zone);
 
 		// Compute config for all devices
 		Map<String, DeviceConfig> configMap = new HashMap<>();
--- a/src/main/java/com/facebook/openwifirrm/DeviceDataManager.java
+++ b/src/main/java/com/facebook/openwifirrm/DeviceDataManager.java
@@ -458,7 +458,7 @@ public class DeviceDataManager {
 	 * Return config (with all config layers applied) for all devices in a given
 	 * zone, or null if not present in the topology.
 	 *
-	 * @param zone the zone that will be looked up for devi es
+	 * @param zone the zone that will be looked up for devices
 	 *
 	 * @return map of serial number to computed config
 	 */
@@ -551,7 +551,7 @@ public class DeviceDataManager {
 	}
 
 	/** Device AP config layer update interface. */
-	public interface LayerConfigFunction {
+	public interface ZoneConfigFunction {
 		/** Update the AP config layer. */
 		void update(Map<String, DeviceConfig> apConfig);
 	}
@@ -562,7 +562,7 @@ public class DeviceDataManager {
 	 * @param fn The function to apply for the configs. The key of the argument is
 	 *        the zone name.
 	 */
-	public void updateZoneConfig(LayerConfigFunction fn) {
+	public void updateZoneConfig(ZoneConfigFunction fn) {
 		Lock l = deviceLayeredConfigLock.writeLock();
 		l.lock();
 		try {
@@ -575,13 +575,19 @@ public class DeviceDataManager {
 		saveDeviceLayeredConfig();
 	}
 
+	/** Device AP config layer update interface. */
+	public interface DeviceApConfigFunction {
+		/** Update the AP config layer. */
+		void update(Map<String, DeviceConfig> apConfig);
+	}
+
 	/**
 	 * Apply updates to the device AP config layer (under a write lock).
 	 *
 	 * @param fn The function to apply for the configs. The key of the argument is
 	 *        the serial number of the device.
 	 */
-	public void updateDeviceApConfig(LayerConfigFunction fn) {
+	public void updateDeviceApConfig(DeviceApConfigFunction fn) {
 		Lock l = deviceLayeredConfigLock.writeLock();
 		l.lock();
 		try {
--- a/src/main/java/com/facebook/openwifirrm/DeviceDataManager.java
+++ b/src/main/java/com/facebook/openwifirrm/DeviceDataManager.java
@@ -454,6 +454,13 @@ public class DeviceDataManager {
 		);
 	}
 
+	/**
+	 * Return the serail number of devices in a given zone.
+	 *
+	 * @param zone the zone that will be looked up for devi es
+	 *
+	 * @return set of serial numbers
+	 */
 	private Set<String> getAllDevicesInZone(String zone) {
 		// Get all devices in zone
 		if (zone == null || zone.isEmpty()) {
@@ -477,6 +484,8 @@ public class DeviceDataManager {
 	/**
 	 * Return config (with all config layers applied) for all devices in a given
 	 * zone, or null if not present in the topology.
+	 *
+	 * @param zone the zone that will be looked up for devi es
 	 *
 	 * @return map of serial number to computed config
 	 */
--- a/src/main/java/com/facebook/openwifirrm/modules/ApiServer.java
+++ b/src/main/java/com/facebook/openwifirrm/modules/ApiServer.java
@@ -14,7 +14,6 @@ import java.net.URISyntaxException;
 import java.net.UnknownHostException;
 import java.time.Instant;
 import java.util.Arrays;
-import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
@@ -43,7 +42,6 @@ import com.facebook.openwifirrm.DeviceTopology;
 import com.facebook.openwifirrm.RRMAlgorithm;
 import com.facebook.openwifirrm.RRMConfig.ModuleConfig.ApiServerParams;
 import com.facebook.openwifirrm.RRMConfig.ServiceConfig;
-import com.facebook.openwifirrm.RRMSchedule;
 import com.facebook.openwifirrm.Utils.LruCache;
 import com.facebook.openwifirrm.VersionProvider;
 import com.facebook.openwifirrm.optimizers.channel.LeastUsedChannelOptimizer;
@@ -57,11 +55,6 @@ import com.facebook.openwifirrm.ucentral.UCentralClient;
 import com.facebook.openwifirrm.ucentral.UCentralUtils;
 import com.facebook.openwifirrm.ucentral.gw.models.SystemInfoResults;
 import com.facebook.openwifirrm.ucentral.gw.models.TokenValidationResult;
-import com.facebook.openwifirrm.ucentral.prov.models.InventoryTag;
-import com.facebook.openwifirrm.ucentral.prov.models.InventoryTagList;
-import com.facebook.openwifirrm.ucentral.prov.models.RRMAlgorithmDetails;
-import com.facebook.openwifirrm.ucentral.prov.models.RRMDetails;
-import com.facebook.openwifirrm.ucentral.prov.models.SerialNumberList;
 import com.facebook.openwifirrm.ucentral.prov.rrm.models.Algorithm;
 import com.facebook.openwifirrm.ucentral.prov.rrm.models.Provider;
 import com.google.gson.Gson;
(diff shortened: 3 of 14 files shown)


src/main/java/com/facebook/openwifirrm/ucentral/prov/models/RrmAlgorithmDetails.java
```diff
@@ -0,0 +1,14 @@
...
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+package com.facebook.openwifirrm.ucentral.prov.models;
+
+public class RrmAlgorithmDetails {
```
elludraon, line 11: You can capitalize this name as `RRMAlgorithmDetails`, see [src/RESTObjects/RESTAPI_ProvObjects.h](https://github.com/Telecominfraproject/wlan-cloud-owprov/pull/55/files#diff-69a248eae759933adffac236c046f0c279e3b65852b5320d2ef043966d0234f3)

src/main/java/com/facebook/openwifirrm/ucentral/prov/models/RrmDetails.java
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+package com.facebook.openwifirrm.ucentral.prov.models;
+
+import java.util.List;
+
+public class RrmDetails {
```
elludraon, line 13: same, `RRMDetails`

src/main/java/com/facebook/openwifirrm/ucentral/prov/models/InventoryConfigApplyResult.java
```diff
@@ -0,0 +1,18 @@
...
+
+package com.facebook.openwifirrm.ucentral.prov.models;
+
+import java.util.List;
+
+public class InventoryConfigApplyResult {
+	public String appliedConfiguration;
+	public List<String> errors;
+	public List<String> warnings;
+	public Integer errorCode;
```
elludraon, line 17: any reason you're using boxed `Integer` instead of primitive `int`? (is the parameter not always filled?)

src/main/java/com/facebook/openwifirrm/DeviceConfig.java
```diff
@@ -30,6 +30,9 @@ public class DeviceConfig {
 	/** The RRM schedule (if enabled), only applied at the zone or network layer */
 	public RRMSchedule schedule;
 
+	/** earameters for the algorithms, mapping algorithm name to string of a list of parameters */
+	public Map<String, String> algorithmsParameters;
+
```
elludraon, line 35: This is already defined in the `RRMSchedule` class, don't duplicate it here: https://github.com/Telecominfraproject/wlan-cloud-rrm/blob/6fb15f777cc8034e9e67fd6bdfd0e75f5aeeac89/src/main/java/com/facebook/openwifirrm/RRMSchedule.java#L31

jwoos: ah missed that, thanks for the catch!

src/main/java/com/facebook/openwifirrm/DeviceDataManager.java
```diff
@@ -454,13 +454,7 @@ public DeviceConfig getDeviceConfig(String serialNumber, String zone) {
...
 	}
 
-	/**
-	 * Return config (with all config layers applied) for all devices in a given
-	 * zone, or null if not present in the topology.
-	 *
-	 * @return map of serial number to computed config
-	 */
-	public Map<String, DeviceConfig> getAllDeviceConfigs(String zone) {
+	private Set<String> getAllDevicesInZone(String zone) {
```
jaspreetsachdev, line 457: Missing function description

src/main/java/com/facebook/openwifirrm/ucentral/prov/models/RRMDetails.java
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+package com.facebook.openwifirrm.ucentral.prov.models;
+
+import java.util.List;
+
+public class RRMDetails {
```
wuweicaifb, line 13: Any reason to wrap `RRMDetailsImpl`?

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
```diff
@@ -591,6 +593,71 @@ public SerialNumberList getProvInventoryForRRM() {
 		}
 	}
 
+	/** Retrieve the RRM config and schedule for a specific AP */
```
wuweicaifb, line 596: Please use our standard format of java doc, i.e. including the args in the java doc.

src/main/java/com/facebook/openwifirrm/DeviceDataManager.java
```diff
@@ -477,6 +471,18 @@ public Map<String, DeviceConfig> getAllDeviceConfigs(String zone) {
...
+		return devicesInZone;
+	}
+
+	/**
+	 * Return config (with all config layers applied) for all devices in a given
+	 * zone, or null if not present in the topology.
+	 *
+	 * @return map of serial number to computed config
+	 */
+	public Map<String, DeviceConfig> getAllDeviceConfigs(String zone) {
```
wuweicaifb, line 483: Include arg in the javadoc.

wuweicaifb (commented):
It LGTM. but a few comments on the javadoc of new functions. Besides, do you have any tests for your update?

src/main/java/com/facebook/openwifirrm/ucentral/prov/models/RRMDetails.java
jwoos: Just following the prov side of things!

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
jwoos: sounds good - got confused because i've seen a lot that only has the comment but will keep in mind for future

jwoos:
> It LGTM. but a few comments on the javadoc of new functions. Besides, do you have any tests for your update?

unfortunately I can only test this manually - the prov sync tests weren't written and I think we were waiting to move to the new schema before writing tests correct me if I'm wrong, @elludraon

wuweicaifb (approved):

src/main/java/com/facebook/openwifirrm/modules/ApiServer.java
```diff
@@ -790,7 +880,11 @@ public String handle(
 				modeler.revalidate();
 
 				// Update scheduler
-				scheduler.syncTriggers();
+
+				// TODO [ZoneBasedRrmScheduling] reeanable after venue based configs
```
zhiqiand, line 884: typo: reenable

src/main/java/com/facebook/openwifirrm/modules/RRMScheduler.java
```diff
@@ -156,11 +169,101 @@ public void shutdown() {
...
+					}
+				} catch (SchedulerException e) {
+					logger.error(
+						"Failed to schedule RRM trigger for device: " +
+							serialNumber,
+						e
+					);
+					continue;
+				}
+				scheduled.add(serialNumber);
```
zhiqiand, line 235: Should we put this line in the try block? Only APs that are successfully scheduled are added.

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
```diff
@@ -591,6 +593,78 @@ public SerialNumberList getProvInventoryForRRM() {
...
+		HttpResponse<String> response =
+			httpGet(String.format("inventory/%s", serial), OWPROV_SERVICE, parameters);
+		if (!response.isSuccess()) {
+			logger.error("Error: {}", response.getBody());
+			return null;
+		}
+		try {
+			InventoryConfigApplyResult result = gson
+				.fromJson(response.getBody(), InventoryConfigApplyResult.class);
+			if (result.warnings != null) {
```
zhiqiand, line 642: Maybe also check if the `warnings` or the `errors` are empty? I see them are defined as list in `InventoryConfigApplyResult` class.

src/main/java/com/facebook/openwifirrm/modules/ProvMonitor.java
```diff
@@ -105,25 +110,41 @@ private void runImpl() {
...
+			if (details == null) {
+				logger
+					.error("Could not fetch RRM details for {}", serialNumber);
+				continue;
+			}
+
+			rrmDetails.put(serialNumber, details);
+		}
+
+		VenueList venueList = client.getProvVenues();
```
zhiqiand, line 133: My question is that, will `getProvVenues` affect fetching RRM details process? If `getProvVenues` failed, is it expected to early exit the method?

zhiqiand (commented):
Thanks for contribution on getting RRM configs! It looks good to me but just some nits. You could just ignore me mark them as resolved if I was wrong

src/main/java/com/facebook/openwifirrm/modules/ApiServer.java
```diff
@@ -424,6 +431,89 @@ private String getOpenApiJson(Request request, Response response) {
 		return Json.pretty(getOpenApi());
 	}
 
+	/** TODO [ZoneBasedRrmScheduling] remove this once we move to venues
```
wuweicaifb, line 434: I think we should place `TODO`s outside of the javadoc, because javadoc is used to generate documents and it's not good to include development TODOs in the documentation. But maybe I'm wrong, feel free to push back. @elludraon

src/main/java/com/facebook/openwifirrm/modules/RRMScheduler.java
jwoos: the loop will continue without getting to this point due to the `continue` on line 233

src/main/java/com/facebook/openwifirrm/modules/ProvMonitor.java
jwoos: So in this implementation, we don't technically need this call. However, the "proper" implementation would be that prov would return RRM schedules per venue, not resolved config per device. So if this fails in that implementation, then it makes sense to exit the method early

src/main/java/com/facebook/openwifirrm/modules/ApiServer.java
jwoos: I need to get used to javadoc, i think it makes sense to separate it.

zhiqiand: There should be some extensions/tools in your IDE to auto-generated a javadoc template for a method. Then you could fill in that template easily. Usually you are only expeced to add method description, @param, @return. For this PR, I ok that if you do it separately.

src/main/java/com/facebook/openwifirrm/modules/ProvMonitor.java
zhiqiand: If it makes sense to exit the method early, maybe putting these lines before fetching RRM details process?

zhiqiand (approved):
LGTM

src/main/java/com/facebook/openwifirrm/DeviceDataManager.java
```diff
@@ -455,12 +455,13 @@ public DeviceConfig getDeviceConfig(String serialNumber, String zone) {
...
-	 * zone, or null if not present in the topology.
+	 * Return the serail number of devices in a given zone.
 	 *
-	 * @return map of serial number to computed config
+	 * @param zone the zone that will be looked up for devi es
+	 *
+	 * @return set of serial numbers
 	 */
-	public Map<String, DeviceConfig> getAllDeviceConfigs(String zone) {
+	private Set<String> getAllDevicesInZone(String zone) {
```
elludraon, line 464: Why did you split this function? `getAllDevicesInZone()` (new function) doesn't look used in this PR. If you do need this, note this function can return null which you need to handle in `getAllDeviceConfigs()` (currently would throw NPE) and mention in `@return` javadoc.

src/main/java/com/facebook/openwifirrm/modules/ApiServer.java
```diff
@@ -424,6 +431,90 @@ private String getOpenApiJson(Request request, Response response) {
 		return Json.pretty(getOpenApi());
 	}
 
+	// TODO [ZoneBasedRrmScheduling] remove this once we move to venues
+	/**
+	 * Calls prov to get a mapping of all RRM enabled devices to their settings
+	 * and updates the device config.
+	 */
+	private void getAndUpdateRRMDetails() {
```
elludraon, line 439: As discussed, when Prov sync is enabled then this isn't needed as we should be blocking topology changes, as well as config changes that specifically modify Prov-managed RRM settings (this is much more annoying to check). You can add some TODOs to the config setters like the one in `/setToplogy` and get to this later.

src/main/java/com/facebook/openwifirrm/modules/RRMScheduler.java
```diff
@@ -178,6 +281,20 @@ public void syncTriggers() {
...
 
+			try {
+				CronExpression.validateExpression(config.schedule.cron);
+			} catch (ParseException e) {
+				logger.error(
+					String.format(
+						"Invalid cron expression (%s) for zone %s",
+						config.schedule.cron,
+						zone
+					) + zone,
```
elludraon, line 292: typo? `+ zone`

src/main/java/com/facebook/openwifirrm/modules/RRMScheduler.java
```diff
@@ -75,6 +78,12 @@ public class RRMScheduler {
 	/** The zones with active triggers scheduled. */
 	private Set<String> scheduledZones;
 
+	/**
+	 * TODO [ZoneBasedRrmScheduling] remove this in favor of zones.
+	 * The devices with active triggers scheduled.
+	 */
+	private Set<String> scheduledDevices;
```
elludraon, line 85: As discussed, this isn't needed or useful - RRM should be done at venue level.

src/main/java/com/facebook/openwifirrm/modules/ProvMonitor.java
```diff
@@ -105,25 +110,41 @@ private void runImpl() {
...
+			RRMDetails details =
+				client.getProvInventoryRrmDetails(serialNumber);
+			if (details == null) {
+				logger
+					.error("Could not fetch RRM details for {}", serialNumber);
+				continue;
+			}
+
+			rrmDetails.put(serialNumber, details);
+		}
```
elludraon, line 131: As discussed, we should assume that RRM config is only set at venue level, and the workaround with current Prov APIs would be to query `RRMDetails` for one (random/first) device in each venue -- you don't need to query all devices. And hopefully `getProvInventoryForRRM()` API still works here to filter RRM-enabled devices only (should test this if you haven't already).

src/main/java/com/facebook/openwifirrm/modules/ProvMonitor.java
```diff
@@ -181,6 +203,40 @@ protected void syncDataToProv(
...
+						logger.error(
+							"No RRM details available for {} even though it has RRM enabled",
+							serialNumber
+						);
+						continue;
+					}
+
+					cfg.schedule = new RRMSchedule();
+					if (details.rrm != null) {
+						cfg.schedule.cron = details.rrm.schedule;
```
elludraon, line 219: As discussed, let's try to convert the schedule from standard cron into Quartz syntax. You can create a public static method in `RRMScheduler` for this, and maybe return null and skip/log error here if the schedule string is invalid.

elludraon (changes requested):

src/main/java/com/facebook/openwifirrm/DeviceDataManager.java
jwoos: I think I was going to use it then I took a different approach, will revert!

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
```diff
@@ -591,6 +593,78 @@ public SerialNumberList getProvInventoryForRRM() {
...
+				"Failed to deserialize to RRMDetails: %s",
+				response.getBody()
+			);
+			logger.error(errMsg, e);
+			return null;
+		}
+	}
+
+	/** Retrieve resolved inventory */
+	public InventoryTagList getProvInventoryGetConfigApplyResult(
```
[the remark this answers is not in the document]
jwoos: Yeah, I was looking at the OpenAPI but you are correct. I'll remove for now.

src/main/java/com/facebook/openwifirrm/DeviceDataManager.java
```diff
@@ -458,10 +458,11 @@ public DeviceConfig getDeviceConfig(String serialNumber, String zone) {
 	 * Return config (with all config layers applied) for all devices in a given
 	 * zone, or null if not present in the topology.
 	 *
+	 * @param zone the zone that will be looked up for devi es
```
elludraon, line 461: typo: "devi es"

you don't need to write `@param` when it's obvious

src/main/java/com/facebook/openwifirrm/DeviceDataManager.java
```diff
@@ -550,13 +551,37 @@ public void setDeviceApConfig(String serialNumber, DeviceConfig apConfig) {
 	}
 
 	/** Device AP config layer update interface. */
-	public interface DeviceApConfigFunction {
+	public interface LayerConfigFunction {
```
elludraon, line 554: Instead of renaming this, can you make a separate interface instead for AP vs. zone? It barely duplicates any code and will make it clearer (ex. when searching). Also, the same interface method wouldn't work for updating the network layer.

src/main/java/com/facebook/openwifirrm/ucentral/prov/models/Venue.java
```diff
@@ -29,7 +29,7 @@ public class Venue {
 	public List<String> devices;
 	public DiGraph topology;
 	public String design;
-	//public String deviceConfiguration;  // TODO wrong in OpenAPI (array?)
+	public List<String> deviceConfiguration;
```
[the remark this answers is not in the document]
elludraon: ^bump

src/main/java/com/facebook/openwifirrm/modules/ProvMonitor.java
```diff
@@ -105,29 +110,58 @@ private void runImpl() {
...
+		// the RRMDetails. Once a proper venue based RRM settings API is available
+		// we should use that instead of doing this.
+		for (InventoryTag tag : inventory.taglist) {
+			rrmDetails.computeIfAbsent(tag.venue, k -> {
+				RRMDetails details =
+					client.getProvInventoryRRMDetails(tag.serialNumber);
+				if (details == null) {
+					logger
+						.error(
+							"Could not fetch RRM details for device {} and venue {}",
```
elludraon, line 131: nit: "**and** venue" -> "**in** venue"

src/main/java/com/facebook/openwifirrm/modules/ProvMonitor.java
```diff
@@ -105,29 +110,58 @@ private void runImpl() {
...
 		SerialNumberList inventoryForRRM = client.getProvInventoryForRRM();
-		VenueList venueList = client.getProvVenues();
 		//EntityList entityList = client.getProvEntities();
 		if (inventory == null || inventoryForRRM == null) {
 			logger.error("Failed to fetch inventory from owprov");
 			return;
 		}
+
+		// fetch the RRM details for each venue
+		Map<String, RRMDetails> rrmDetails = new HashMap<String, RRMDetails>();
```
elludraon, line 120: nit: you can omit the `<>` generic arguments in the instantiation when it can be inferred:
```java
Map<String, RRMDetails> rrmDetails = new HashMap<>();
```

src/main/java/com/facebook/openwifirrm/modules/ProvMonitor.java
```diff
@@ -160,8 +194,79 @@ protected void syncDataToProv(
...
+							"Venue name {} is not found in id mapping",
+							venue
+						);
+						continue;
+					}
+
+					DeviceConfig cfg = configMap
+						.computeIfAbsent(zone, k -> new DeviceConfig());
+					cfg.enableRRM =
+						cfg.enableConfig = cfg.enableWifiScan = true;
```
elludraon, line 221: you don't need to do this here because the block below syncing device configs will take care of it -- it shouldn't be in two places

src/main/java/com/facebook/openwifirrm/modules/ProvMonitor.java
```diff
@@ -160,8 +194,79 @@ protected void syncDataToProv(
...
+					if (details == null) {
+						logger.error(
+							"No RRM details available for zone {} even though it has RRM enabled",
+							zone
+						);
+						continue;
+					}
+
+					cfg.schedule = new RRMSchedule();
+					if (details.rrm != null) {
```
elludraon, line 234: Both of these checks should be done before modifying `configMap` (don't add non-null but empty `cfg.schedule` -- null is OK)

src/main/java/com/facebook/openwifirrm/modules/ProvMonitor.java
```diff
@@ -160,8 +194,79 @@ protected void syncDataToProv(
 			topo.values().stream().mapToInt(x -> x.size()).sum()
 		);
 
+		// Sync zone configs
+		deviceDataManager.updateZoneConfig(
+			configMap -> {
+				for (
+					Map.Entry<String, RRMDetails> entry : rrmDetails.entrySet()
+				) {
```
elludraon, line 202: You need to handle the case where some venue RRM config got newly wiped in Prov (and all venue devices were also deleted). So, before this loop, you should wipe zone config for all zones in `venueList.venues`.

This will also save you from needing to null out the venue config in the logic below if empty in Prov data, which you aren't doing now.

src/main/java/com/facebook/openwifirrm/modules/ProvMonitor.java
```diff
@@ -105,29 +110,58 @@ private void runImpl() {
...
 			return;
 		}
+
+		// fetch the RRM details for each venue
+		Map<String, RRMDetails> rrmDetails = new HashMap<String, RRMDetails>();
+		// TODO this currently chooses the first AP encountered per venue to fetch
+		// the RRMDetails. Once a proper venue based RRM settings API is available
+		// we should use that instead of doing this.
+		for (InventoryTag tag : inventory.taglist) {
+			rrmDetails.computeIfAbsent(tag.venue, k -> {
```
elludraon, line 125: Skip if `tag.venue.isEmpty()` (a device may not have a venue if it got auto-added but not yet configured)

src/main/java/com/facebook/openwifirrm/modules/ProvMonitor.java
elludraon: ^ I would follow this suggestion and move the `VenueList` block up.

src/main/java/com/facebook/openwifirrm/modules/ProvMonitor.java
```diff
@@ -160,8 +194,79 @@ protected void syncDataToProv(
...
+							) {
+								cfg.schedule.algorithms.add(
+									RRMAlgorithm.parse(
+										algorithmDetails.name,
+										algorithmDetails.parameters
+									)
+								);
+							}
+						}
+						logger.info("Set RRM settings for zone {}", zone);
```
elludraon, line 260: I would create a sub-method in this class which builds `RRMSchedule` from `RRMDetails` input, or null upon failure.

src/main/java/com/facebook/openwifirrm/modules/RRMScheduler.java
```diff
@@ -95,6 +97,60 @@ public void execute(JobExecutionContext context)
...
+		}
+
+		String[] split = linuxCron.split(" ");
+		if (split.length < 6 || split.length > 7) {
+			// Quartz cron MUST have at least 6 and no more than 7 fields
+			return null;
+		}
+
+		String dayOfMonth = split[3];
+		String dayOfWeek = split[5];
```
elludraon, line 126: nit: before this, also declare
```
final int MONTH_INDEX = 3;
final int WEEK_INDEX = 5;
```
so that you don't need to hardcode the indices multiple times (you're using them below)

src/main/java/com/facebook/openwifirrm/modules/ProvMonitor.java
```diff
@@ -160,8 +194,79 @@ protected void syncDataToProv(
...
+							for (
+								RRMAlgorithmDetails algorithmDetails : details.rrm.algorithms
+							) {
+								cfg.schedule.algorithms.add(
+									RRMAlgorithm.parse(
+										algorithmDetails.name,
+										algorithmDetails.parameters
+									)
+								);
+							}
```
elludraon, line 258: nit: use streams for this (stream->map->collect replacing simple "for" loop)
```java
cfg.schedule.algorithms = details.rrm.algorithms.stream().map(details ->
    RRMAlgorithm.parse(details.name, details.parameters)
).collect(Collectors.toList());

```

elludraon (changes requested):

elludraon (approved):
Few last comments

PR #68, opened 2022-09-09
stats aggregation and reports to json format string

zhiqiand:
### Summary
This PR Aggregates the stats info per AP so we can have a clear picture of the current system. 
- Aggregation condition: 
The following condition is true then two stats results need to be aggregated
channel, channel width, and tx_power under "radios" match
- Aggregation rules:
The current data structure is AP has many clients under “associations”. 
For a client with the same mac address, average/list all the stats over time.

### Implementation 
- The mainly functional method is `ModelerUtils.getAggregatedStates`, which group States by given `aggregatedFields`(RSSI, MCS...) and `aggregatedKeys`(channel, channel width, tx_power...). Only States that are non-obsolete are to be aggregated. The aggregation map is a large nested map. The following picture shows the what the structure is.

IMG_7ED79495C2BE-1

- `ModelerUtils.aggregatedStatesToJson` serializes the aggregated map to a Json String. There are two formats to represent the values : list of values(a list of all the non-obsolete stats) or single value(take mean/median... of all the stats). A custom serializer is used to switch over between the two formats.
- Add `getList` in `Aggregator`, which supports the json serializer to give a list view.

### Questions
- A new field `latestStates` which is a list of most recent X States (X = 10...) is added, whereas `latestState` is the most recent State. Not sure if [PR#61](https://github.com/Telecominfraproject/wlan-cloud-rrm/pull/61) get affected.
- Please let me know if the keys in the nested aggregation map are redundant or need to be flattened.
- Not confident with the wording in javadoc. Please check the method description carefully.

### DIFF
--- a/src/main/java/com/facebook/openwifirrm/aggregators/Aggregator.java
+++ b/src/main/java/com/facebook/openwifirrm/aggregators/Aggregator.java
@@ -8,6 +8,8 @@
 
 package com.facebook.openwifirrm.aggregators;
 
+import java.util.List;
+
 /**
  * Aggregates added values into one "aggregate" measure.
  *
@@ -19,6 +21,9 @@ public interface Aggregator<T> {
 
 	/** Returns the aggregate measure of all added values. */
 	T getAggregate();
+
+	/** Returns a list of all added values. */
+	List<T> getList();
 
 	/** Returns the number of values that are aggregated. */
 	long getCount();
--- a/src/main/java/com/facebook/openwifirrm/aggregators/MeanAggregator.java
+++ b/src/main/java/com/facebook/openwifirrm/aggregators/MeanAggregator.java
@@ -8,16 +8,30 @@
 
 package com.facebook.openwifirrm.aggregators;
 
+import java.util.LinkedList;
+import java.util.List;
+
 /**
  * Tracks the mean of all added values. If no values are added, the mean is 0.
  */
 public class MeanAggregator implements Aggregator<Double> {
-
+	protected List<Double> valueList = new LinkedList<>();
 	protected double mean = 0;
 	protected long count = 0;
 
+	/** Constructor with no args */
+	public MeanAggregator() {}
+
+	/** Constructor with list of values*/
+	public MeanAggregator(List<Double> values) {
+		for (Double value : values) {
+			addValue(value);
+		}
+	}
+
 	@Override
 	public void addValue(Double value) {
+		valueList.add(value);
 		mean = ((double) count / (count + 1)) * mean + (value / (count + 1));
 		count++;
 	}
@@ -26,10 +40,14 @@ public class MeanAggregator implements Aggregator<Double> {
 	public Double getAggregate() { return mean; }
 
 	@Override
+	public List<Double> getList() { return valueList; }
+
+	@Override
 	public long getCount() { return count; }
 
 	@Override
 	public void reset() {
+		valueList = new LinkedList<>();
 		mean = 0;
 		count = 0;
 	}
--- a/src/main/java/com/facebook/openwifirrm/modules/Modeler.java
+++ b/src/main/java/com/facebook/openwifirrm/modules/Modeler.java
@@ -8,6 +8,8 @@
 
 package com.facebook.openwifirrm.modules;
 
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
@@ -94,6 +96,10 @@ public class Modeler implements Runnable {
 			new ConcurrentHashMap<>();
 
 		/** List of latest state per device. */
+		public Map<String, List<State>> latestStates =
+			new ConcurrentHashMap<>();
+
+		/** The most latest state per device.*/
 		public Map<String, State> latestState = new ConcurrentHashMap<>();
 
 		/** List of radio info per device. */
@@ -268,6 +274,10 @@ public class Modeler implements Runnable {
 				try {
 					State stateModel = gson.fromJson(state, State.class);
 					dataModel.latestState.put(device.serialNumber, stateModel);
+					dataModel.latestStates.put(
+						device.serialNumber,
+						new ArrayList<>(Arrays.asList(stateModel))
+					);
 					logger.debug(
 						"Device {}: added initial state from uCentralGw",
 						device.serialNumber
@@ -301,6 +311,12 @@ public class Modeler implements Runnable {
 						State stateModel = gson.fromJson(state, State.class);
 						dataModel.latestState
 							.put(record.serialNumber, stateModel);
+						dataModel.latestStates
+							.computeIfAbsent(
+								record.serialNumber,
+								k -> new ArrayList<>()
+							)
+							.add(stateModel);
 						stateUpdates.add(record.serialNumber);
 					} catch (JsonSyntaxException e) {
 						logger.error(
@@ -424,6 +440,12 @@ public class Modeler implements Runnable {
 		}
 		if (
 			dataModel.latestState.entrySet()
+				.removeIf(e -> !isRRMEnabled(e.getKey()))
+		) {
+			logger.debug("Removed some state entries from data model");
+		}
+		if (
+			dataModel.latestStates.entrySet()
 				.removeIf(e -> !isRRMEnabled(e.getKey()))
 		) {
 			logger.debug("Removed some state entries from data model");
--- a/src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
+++ b/src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
@@ -8,7 +8,9 @@
 
 package com.facebook.openwifirrm.modules;
 
+import java.lang.reflect.Type;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -22,13 +24,29 @@ import com.facebook.openwifirrm.aggregators.Aggregator;
 import com.facebook.openwifirrm.aggregators.MeanAggregator;
 import com.facebook.openwifirrm.modules.Modeler.DataModel;
 import com.facebook.openwifirrm.ucentral.UCentralUtils.WifiScanEntry;
+import com.facebook.openwifirrm.ucentral.models.State;
+import com.facebook.openwifirrm.ucentral.models.State.Interface;
+import com.facebook.openwifirrm.ucentral.models.State.Interface.SSID;
+import com.facebook.openwifirrm.ucentral.models.State.Interface.SSID.Association;
 import com.facebook.openwifirrm.ucentral.operationelement.HTOperationElement;
 import com.facebook.openwifirrm.ucentral.operationelement.VHTOperationElement;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import com.google.gson.JsonArray;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonPrimitive;
+import com.google.gson.JsonSerializationContext;
+import com.google.gson.JsonSerializer;
+import com.google.gson.reflect.TypeToken;
 
 /**
  * Modeler utilities.
  */
 public class ModelerUtils {
+	public enum AggregationType { LIST, MEAN };
+
 	private static final Logger logger =
 		LoggerFactory.getLogger(ModelerUtils.class);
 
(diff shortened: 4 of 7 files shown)


src/main/java/com/facebook/openwifirrm/modules/Modeler.java
```diff
@@ -94,6 +96,10 @@ public static class DataModel {
 			new ConcurrentHashMap<>();
 
 		/** List of latest state per device. */
+		public Map<String, List<State>> latestStates =
```
wuweicaifb, line 99: Please update javadoc to distinguish `latestStates` from `latestState`

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -381,4 +399,269 @@ public static Map<String, Map<String, WifiScanEntry>> getAggregatedWifiScans(
...
+					 * return a JsonArray of the value list.
+					 */
+					@Override
+					public JsonElement serialize(
+						MeanAggregator agg,
+						java.lang.reflect.Type typeOfSrc,
+						JsonSerializationContext context
+					) {
+						JsonArray aggJson = new JsonArray();
+						for (Double value : agg.getList()) {
```
wuweicaifb, line 435: You can use `Gson` to convert an `ArrayList` to `JsonArray`.

src/main/java/com/facebook/openwifirrm/aggregators/MeanAggregator.java
```diff
@@ -8,28 +8,46 @@
...
+	public MeanAggregator(List<Double> values) {
+		for (Double value : values) {
+			addValue(value);
+		}
+	}
+
 	@Override
 	public void addValue(Double value) {
+		valueList.add(value);
 		mean = ((double) count / (count + 1)) * mean + (value / (count + 1));
```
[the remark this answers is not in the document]
wuweicaifb: Or you can even add a new private double `valueSum` to store the sum of the values.

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
zhiqiand: I guess it might be the same if I initial it as a JsonArray or ArrayList?

src/test/java/com/facebook/openwifirrm/aggregators/MeanAggregatorTest.java
```diff
@@ -22,35 +25,51 @@ public void testEmptyAndNonEmptyAndReset() {
 		// default mean is 0
 		assertEquals(0, agg.getAggregate(), eps);
 		assertEquals(0, agg.getCount());
+		assertEquals(new LinkedList<>(), agg.getList());
 
 		// adding 0 (the mean) does not change the mean
 		agg.addValue(0.0);
 		assertEquals(0, agg.getAggregate(), eps);
 		assertEquals(1, agg.getCount());
+		assertEquals(new LinkedList<>(Arrays.asList(0.0)), agg.getList());
```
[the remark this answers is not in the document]
zhiqiand: Though the content and assertion result are the same, I put `LinkedList` here emphasizing my purpose is to checking the order.

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
zhiqiand: Where to put the javadoc? I assumed the description in line 91 and line 102 are enough. Open to any suggestions

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
```diff
@@ -94,6 +96,10 @@ public static class DataModel {
...
+			new ConcurrentHashMap<>();
+
+		/** The most latest state per device.*/
 		public Map<String, State> latestState = new ConcurrentHashMap<>();
```
jwoos, line 103: wondering if it's cleaner to just have one state mapping, and just reference the last element in the list vs having two separate state mapping. This would also necessitate that the list become thread safe as well though so I'm not sure :)

src/main/java/com/facebook/openwifirrm/aggregators/MeanAggregator.java
```diff
@@ -8,28 +8,46 @@
 
 package com.facebook.openwifirrm.aggregators;
 
+import java.util.LinkedList;
+import java.util.List;
+
 /**
  * Tracks the mean of all added values. If no values are added, the mean is 0.
  */
 public class MeanAggregator implements Aggregator<Double> {
```
jwoos, line 17: General question, do we only care about average for the whole duration? What about mean of a time frame?

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -381,4 +399,269 @@ public static Map<String, Map<String, WifiScanEntry>> getAggregatedWifiScans(
...
 	}
+
+	/** This method converts the agggregated map to a Json String.
+	 *
+	 * @param aggregatedMap 	Aggregated map
+	 * @param aggType 			Determines how to aggregate the statistics.
+	 * @return String			A Json String representation of the aggregated map.
+	 */
+	public static String aggregatedStatesToJson(
+		Map<String, Map<String, Map<String, Map<JsonNode, Map<String, MeanAggregator>>>>> aggregatedMap,
```
jwoos, line 410: 😬 

it makes sense but a few things I don't like about this:
- too nested
- `String` doesn't give enough information about what kind of string it actually is

I won't block this because I can't think of a better representation off the top of my head though

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
zhiqiand: That's also one of my questions. Know that you also use the `latestState` filed in #61, would you prefer only keeping `latestStates`? The `latestStates` will keep all the States in the Kafka record without any time frame. Do you have any suggestions on thread-safe?

src/main/java/com/facebook/openwifirrm/aggregators/MeanAggregator.java
zhiqiand: I guess now we only care about average for the whole duration. RSSI, MCS are all discrete type. `MeanAggregator` operates in a given time frame(That is, time frame is not a parameter in the calculation). I guess only when the variables are continuous we need to calculate mean of a time frame, but all the continuous type are decomposed in records.

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
jwoos: I think it makes sense to only keep one, since one of them would essentially be redundant data. 

For thread safety, I'm not so sure since I'm still pretty new to Java - not sure if there's a concurrent list

src/test/java/com/facebook/openwifirrm/aggregators/MeanAggregatorTest.java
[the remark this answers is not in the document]
wuweicaifb: You are able to check the order with `ArrayList` too.

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
wuweicaifb: I think you are converting an `ArraryList` or `LinkedList` to a `JsonArray` here, aren't you? I believe `Gson` have some methods to deal with it directly.

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
wuweicaifb: A nested/chained keys like that also cost much hash lookup time. Based on whether we really want to chain these keys or just want a combination of the keys, we can use different methods. I'd suggest a class with those keys and implement `hashCode` and `equals`, and then use that class as the key if a chained set of keys is not really needed.

zhiqiand: Thanks for comments! I will wait for more inputs from the system team

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -381,4 +399,269 @@ public static Map<String, Map<String, WifiScanEntry>> getAggregatedWifiScans(
...
+				obsoletionPeriodMs,
+				System.currentTimeMillis(),
+				aggregatedKeys,
+				aggregatedFields
+			);
+		} catch (
+			IllegalArgumentException | IllegalAccessException
+			| NoSuchFieldException | SecurityException e
+		) {
+			e.printStackTrace();
```
elludraon, line 526: `e.printStackTrace()` should almost never be present in production code - use `logger.error("description", e)`

elludraon (changes requested):
I am having a really hard time understanding this PR... the goal is to aggregate stats, so shouldn't the return value be the same as the input, i.e. one `State` per device (serial number)?

There are too many methods with similar and long names/arguments. Why are the JSON conversion functions needed?

src/test/java/com/facebook/openwifirrm/optimizers/TestUtils.java
```diff
@@ -544,7 +636,40 @@ public static State createState(
...
+
+	/**
+	 * Create a device state object with one radio.
+	 *
+	 * @param channel channel number
+	 * @param channelWidth channel width in MHz
+	 * @param txPower tx power in dBm
+	 * @param bssid bssid
+	 * @param clientRssis array of client RSSIs
+	 * @param localtime local time
```
RockyMandayam2, line 672: Please specify the units, etc. for local time

src/test/java/com/facebook/openwifirrm/optimizers/TestUtils.java
```diff
@@ -632,7 +803,50 @@ public static State createState(
...
+	 * @param clientRssisA array of client RSSIs for channelA
+	 * @param channelB channel number
+	 * @param channelWidthB channel width (MHz) of channelB
+	 * @param txPowerB tx power for channelB
+	 * @param bssidB bssid for radio on channelB
+	 * @param clientRssisB array of client RSSIs for channelB
+	 * @param localtime local time for the State
+	 * @return the state of an AP with two radios
+	 */
+	public static State createState(
```
RockyMandayam2, line 827: Maybe rename to `createDualRadioState`?

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -381,4 +399,269 @@ public static Map<String, Map<String, WifiScanEntry>> getAggregatedWifiScans(
...
+						 * clientsToStats is the mapping from the station to aggregated statistis.
+						 */
+						Map<String, Map<JsonNode, Map<String, MeanAggregator>>> clientToStats =
+							bssidToStats.computeIfAbsent(
+								ssid.bssid,
+								k -> new HashMap<>()
+							);
+
+						/**
+						 * aggregatedKey is used to Determine if two associations should be aggregated.
```
RockyMandayam2, line 620: "Determine" -> "determine"

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -381,4 +399,269 @@ public static Map<String, Map<String, WifiScanEntry>> getAggregatedWifiScans(
...
+						 * It is a JsonNode type with a mapping between channel/channel width/txpower
+						 * and the related values. The following condition is true then two stats results
```
RockyMandayam2, line 622: I think it is more clear to say "If two radios match in channel, channel widths, and txpower, then the two stats results need to be aggregated."

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
RockyMandayam2: I am going to think about this too, I haven't delved into the details yet...

RockyMandayam2 (changes requested):
I left some comments that have less to do with the core code changes, but I just wanted to give at least some comments earlier. I will do another review once I understand all the changes.

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
pohanhf: Sorry that I think I should have specified it clearly. Both need to be kept. We use the latest one for the optimization algorithm and we use the aggregated one for the root cause analysis. The purpose of this one is a little bit different than the wifiscan aggregation.

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
pohanhf: So I think the nested map can be flatten a little bit. A (bssid, station)-pair defines a link. 
For example, 
serial_number: {
     association: {
          bssid: xxxx,
          station: yyyy,
          key: {
               channel: z, 
               channel_width: e,
               tx_power: r
          },
          stats: {
               mcs: [d1, d2, d3],
               rssi: [f1, f2, f3]
          }
     }
}
Let me know what you guys think about this. Thanks.

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
RockyMandayam2: @pohanhf I think this thread is more about whether we need two variables. We will still have the latest state as the last element in `latestStates` but Jeffrey and I are saying that we don't need a separate `latestState` variable (we can instead to `latestStates.get(0)`

pohanhf: Ah got it. Thanks for the clarification. :)

src/main/java/com/facebook/openwifirrm/aggregators/Aggregator.java
```diff
@@ -20,6 +22,9 @@
 	/** Returns the aggregate measure of all added values. */
 	T getAggregate();
 
+	/** Returns a list of all added values. */
+	List<T> getList();
```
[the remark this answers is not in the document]
zhiqiand: Ah, the motivation of `getList()` call is that sometimes the system team they want a list view of the stats not the mean/median.

src/main/java/com/facebook/openwifirrm/aggregators/MeanAggregator.java
```diff
@@ -8,28 +8,46 @@
...
-
+	protected List<Double> valueList = new LinkedList<>();
```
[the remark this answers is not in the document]
zhiqiand: Do you have other suggestions that we not only print a single value but also support list view?

src/main/java/com/facebook/openwifirrm/aggregators/Aggregator.java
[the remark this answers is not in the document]
RockyMandayam2: Wifi scans have a buffer size (only store up to 10 most recent ones I think). Another comment on this PR mentions to implement a buffer for stats. Therefore, if we _do_ decide to do a `getList()` here, at will at least be indirectly constrained by that... I'll leave it for you, Jeffrey, and WSEs to decide how to proceed.

PR #71, opened 2022-09-14
Minor cleanup in TPC classes

RockyMandayam2:
Updated some comments and removed some unused imports

### DIFF
--- a/src/main/java/com/facebook/openwifirrm/optimizers/tpc/TPC.java
+++ b/src/main/java/com/facebook/openwifirrm/optimizers/tpc/TPC.java
@@ -13,19 +13,18 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.TreeMap;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
-import java.util.List;
-import java.util.Map;
-import java.util.TreeMap;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import com.facebook.openwifirrm.DeviceConfig;
 import com.facebook.openwifirrm.DeviceDataManager;
 import com.facebook.openwifirrm.modules.ConfigManager;
 import com.facebook.openwifirrm.modules.Modeler.DataModel;
 import com.facebook.openwifirrm.ucentral.models.State;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /**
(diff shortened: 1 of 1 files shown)


paulvarkey (approved):
Generally, good. Is there a github lint action that is already setup for this repo, that would automate some styling?

wuweicaifb:
@elludraon added spotless as the formatter in #49, but it's not a github action. We need to run it locally before committing code.

src/main/java/com/facebook/openwifirrm/optimizers/tpc/TPC.java
```diff
@@ -84,25 +83,25 @@ public TPC(
...
-	 * Update the tx power choices based on user and allowed channels from deviceConfig
+	 * Determine the new tx power choices based on user and allowed channels from deviceConfig.
+	 *
 	 * @param band the operational band
-	 * @param serialNumber the device
-	 * @param txPowerChoices the available tx powers of the device
-	 * @return the updated tx powers of the device
+	 * @param serialNumber the AP's serial number
+	 * @param txPowerChoices the AP's available tx powers
+	 * @return the AP's updated tx powers
```
elludraon, line 91: this was intentional, we refer to "device" in the comments because that is what the serial number represents (it's all over the codebase)

src/main/java/com/facebook/openwifirrm/optimizers/tpc/TPC.java
```diff
@@ -84,25 +83,25 @@ public TPC(
...
-			new ArrayList<>(txPowerChoices);
+		List<Integer> newTxPowerChoices = new ArrayList<>(txPowerChoices);
 
-		// Update the available tx powers based on user tx powers or allowed tx powers
 		DeviceConfig deviceCfg = deviceConfigs.get(serialNumber);
 		if (deviceCfg == null) {
 			return newTxPowerChoices;
 		}
+
 		if (
```
elludraon, line 105: This was a general code style choice where we write "paragraphs" (code blocks with no blank lines in between) preceded by a comment. Prefer to keep the comments because it's faster to skim than code.

src/main/java/com/facebook/openwifirrm/optimizers/tpc/TPC.java
RockyMandayam2: We discussed - I'll keep it device for now. Maybe I'll update all references later though.

elludraon (approved):

PR #72, opened 2022-09-14
Add mockito as a dependency

jwoos:
Signed-off-by: Jun Woo Shin <jwoos@fb.com>

Added mockito as our mocking framework of choice.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -227,5 +227,10 @@
       <artifactId>quartz</artifactId>
       <version>2.3.2</version>
     </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -227,5 +227,10 @@
       <artifactId>quartz</artifactId>
       <version>2.3.2</version>
     </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <version>4.8.0</version>
+    </dependency>
```
elludraon, line 234: If we decide to use this, I think this should go in "test" scope?
```xml
<scope>test</scope>
```

PR #73, opened 2022-09-14
Measurement based ap ap tpc test fix

RockyMandayam2:
- Replace `createModel` into `createModelSingleBand` and `createModelDualBand`. Some of the tests earlier were using `createModel` (which did dual band) when they really only should have been single band (though this didn't affect the test).
- "synced" state and status (e.g., if an AP has a 5G radio in state, it should have it in status, and vice versa)
- instead of using "device B" to test what happens when an AP does not have a 5G radio, use device C. This is done to make it consistent with the existing wifiscans which show device B in the 5G band

### DIFF
--- a/src/test/java/com/facebook/openwifirrm/optimizers/TestUtils.java
+++ b/src/test/java/com/facebook/openwifirrm/optimizers/TestUtils.java
@@ -141,6 +141,26 @@ public class TestUtils {
 	}
 
 	/**
+	 * Create an array with one radio info entry with the given tx power and
+	 * channel.
+	 */
+	public static JsonArray createDeviceStatusSingleBand(
+		int channel,
+		int txPower2G
+	) {
+		JsonArray jsonList = new JsonArray();
+		jsonList.add(
+			createDeviceStatusRadioObject(
+				UCentralUtils.getBandFromChannel(channel),
+				channel,
+				DEFAULT_CHANNEL_WIDTH,
+				txPower2G
+			)
+		);
+		return jsonList;
+	}
+
+	/**
 	 * Create an array with two radio info entries (2G and 5G), with the given
 	 * tx powers and channels.
 	 */
--- a/src/test/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPCTest.java
+++ b/src/test/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPCTest.java
@@ -33,7 +33,6 @@ import com.facebook.openwifirrm.optimizers.TestUtils;
 import com.facebook.openwifirrm.ucentral.UCentralConstants;
 import com.facebook.openwifirrm.ucentral.UCentralUtils;
 import com.facebook.openwifirrm.ucentral.WifiScanEntry;
-import com.facebook.openwifirrm.ucentral.models.State;
 
 @TestMethodOrder(OrderAnnotation.class)
 public class MeasurementBasedApApTPCTest {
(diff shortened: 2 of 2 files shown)


src/test/java/com/facebook/openwifirrm/optimizers/TestUtils.java
```diff
@@ -140,6 +140,22 @@ public static JsonArray createDeviceStatus(List<String> bands) {
 		return jsonList;
 	}
 
+	public static JsonArray createDeviceStatusSingleBand(
```
wuweicaifb, line 143: Please add Java doc for this new method

src/test/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPCTest.java
```diff
@@ -78,13 +78,60 @@ private static DeviceDataManager createDeviceDataManager() {
...
+
 	/**
 	 * Creates a data model with 3 devices. All are at max_tx_power, which
 	 * represents the first step in greedy TPC.
 	 *
 	 * @return a data model
 	 */
-	private static DataModel createModel() {
+	private static DataModel createModelDualBand() {
 		DataModel model = new DataModel();
```
wuweicaifb, line 135: I think dual band means a band with two channels in it, rather than an AP has two bands. Just want to make sure we are using the correct terminology. @pohanhf

src/test/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPCTest.java
```diff
@@ -500,14 +550,14 @@ void testComputeTxPowerMapMultiBand() {
 			txPowerMap.get(DEVICE_C).get(UCentralConstants.BAND_2G)
 		);
 
-		// test 5G band
+		// test 5G band (only device A and device B operate in 5G)
```
wuweicaifb, line 553: I think we should test the case that:

1. A device operates in 5G
2. A device does not operate in 5G because 5G band is not available in radio state.
3. A device does not operate in 5G due to missing of wifiscan, which was tested before my PR on the per channel TPC.

FYI, we are not supposed to rely on status to determine which band to operate in, but should rely on state. @jwoos is working on fixing it. We may have further work to distinguish and rename state vs. status.

src/test/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPCTest.java
RockyMandayam2: Dual band means two separate bands, so this is fine

src/test/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPCTest.java
RockyMandayam2: Ok, I'll work on this. For now, I'll keep state and status consistent in tests that I write.

src/test/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPCTest.java
```diff
@@ -78,41 +78,101 @@ private static DeviceDataManager createDeviceDataManager() {
...
+		model.latestDeviceStatus.put(
+			DEVICE_B,
+			TestUtils
+				.createDeviceStatusSingleBand(channel, MAX_TX_POWER)
+		);
+		model.latestDeviceStatus.put(
+			DEVICE_C,
+			TestUtils
+				.createDeviceStatusSingleBand(channel, MAX_TX_POWER)
+		);
```
paulvarkey, line 131:
```suggestion
                 List<String> bssids = Arrays.asList(BSSID_A, BSSID_B, BSSID_C);
                 List<String> devices = Arrays.asList(DEVICE_A, DEVICE_B, DEVICE_C);
                 for (int i = 0; i < devices.size(); i++) { 
                     String device = devices.get(i);
                     String bssid= bssids.get(i);
                     model.latestState.put(device, TestUtils.createState(
			channel,
			DEFAULT_CHANNEL_WIDTH,
			MAX_TX_POWER,
			bssid
		     ));
		     model.latestDeviceStatus.put(
			device,
			TestUtils
				.createDeviceStatusSingleBand(channel, MAX_TX_POWER)
		     );
                }
```

paulvarkey: possibly easier to read and less code?

src/test/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPCTest.java
```diff
@@ -78,41 +78,101 @@ private static DeviceDataManager createDeviceDataManager() {
...
-	private static DataModel createModel() {
+	private static DataModel createModelDualBand() {
 		DataModel model = new DataModel();
 
+		final int channel2G =
+			UCentralUtils.LOWER_CHANNEL_LIMIT.get(UCentralConstants.BAND_2G);
+		final int channel5G =
+			UCentralUtils.LOWER_CHANNEL_LIMIT.get(UCentralConstants.BAND_5G);
+
 		State stateA = TestUtils.createState(
```
paulvarkey, line 150: if you like the suggestion above (purely subjective) -- you may do the same code pattern here

paulvarkey (commented):
Mostly code succinctness suggestion -- but, your style may be different

src/test/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPCTest.java
RockyMandayam2: Good suggestion, thank you.

src/test/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPCTest.java
RockyMandayam2: I do like it, I will do it here too.

paulvarkey (approved):
Approving from my end. The other comments are still outstanding.

src/test/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPCTest.java
```diff
@@ -78,65 +77,86 @@ private static DeviceDataManager createDeviceDataManager() {
 		return deviceDataManager;
 	}
 
+	/**
+	 *
```
elludraon, line 81: delete leading blank line

src/test/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPCTest.java
```diff
@@ -78,65 +77,86 @@ private static DeviceDataManager createDeviceDataManager() {
...
+	 * Creates a data model with 3 devices in only the given band.
+	 * All are at max_tx_power, which represents the first step in greedy TPC.
+	 *
+	 * @param band band (e.g., "2G")
+	 * @return a data model
```
elludraon, line 86: nit: omit `@return` when it doesn't add any info (for example, include it if the method can return null in some cases)

wuweicaifb (approved):

PR #76, opened 2022-09-16
Make AP-AP TPC algorithm use tx power from statistics and fix TPC application to correct band

jwoos:
# Get tx power from statistics
It was possible that the configuration would not have the tx power, which would cause the algorithm to not run. However, with this change tx power will be sourced from statistics instead which should always have tx power. 

- Typed `State`'s `Radio` object. This required changes across a number of files.
  - channel width is string, why???
- Changed `latestDeviceStatus` to `latestDeviceStatusRadios`. This is much clearer since the previous name was confused with `latestState`

# Apply TPC to the correct SSID
Previously, we just took the first SSID of the first interface to run TPC on. This is wrong as the first SSID won't necessarily match up with the band requested (i.e. working on 2.4 but 5 could be retrieved and vice versa). Loop through the interfaces as well as SSID in order to find the correct SSID for the band.

### DIFF
--- a/src/main/java/com/facebook/openwifirrm/optimizers/channel/ChannelOptimizer.java
+++ b/src/main/java/com/facebook/openwifirrm/optimizers/channel/ChannelOptimizer.java
@@ -378,8 +378,13 @@ public abstract class ChannelOptimizer {
 			int tempChannel = state.radios[radioIndex].channel;
 			if (UCentralUtils.isChannelInBand(tempChannel, band)) {
 				currentChannel = tempChannel;
+				// treat as two separate 80MHz channel and only assign to one
+				// TODO: support 80p80 properly
 				Integer parsedChannelWidth = UCentralUtils
-					.parseChannelWidth(state.radios[radioIndex].channel_width);
+					.parseChannelWidth(
+						state.radios[radioIndex].channel_width,
+						true
+					);
 				if (parsedChannelWidth != null) {
 					currentChannelWidth = parsedChannelWidth;
 					break;
--- a/src/main/java/com/facebook/openwifirrm/optimizers/channel/ChannelOptimizer.java
+++ b/src/main/java/com/facebook/openwifirrm/optimizers/channel/ChannelOptimizer.java
@@ -375,14 +375,16 @@ public abstract class ChannelOptimizer {
 			radioIndex < state.radios.length;
 			radioIndex++
 		) {
-			int tempChannel = state.radios[radioIndex]
-				.get("channel")
-				.getAsInt();
+			int tempChannel = state.radios[radioIndex].channel;
 			if (UCentralUtils.isChannelInBand(tempChannel, band)) {
 				currentChannel = tempChannel;
-				currentChannelWidth = state.radios[radioIndex]
-					.get("channel_width")
-					.getAsInt();
+				try {
+					currentChannelWidth = Integer
+						.parseInt(state.radios[radioIndex].channel_width);
+				} catch (NumberFormatException e) {
+					logger.error("Invalid channel width", e);
+					continue;
+				}
 				break;
 			}
 		}
--- a/src/main/java/com/facebook/openwifirrm/optimizers/channel/ChannelOptimizer.java
+++ b/src/main/java/com/facebook/openwifirrm/optimizers/channel/ChannelOptimizer.java
@@ -378,14 +378,18 @@ public abstract class ChannelOptimizer {
 			int tempChannel = state.radios[radioIndex].channel;
 			if (UCentralUtils.isChannelInBand(tempChannel, band)) {
 				currentChannel = tempChannel;
-				try {
-					currentChannelWidth = Integer
-						.parseInt(state.radios[radioIndex].channel_width);
-				} catch (NumberFormatException e) {
-					logger.error("Invalid channel width", e);
-					continue;
+				Integer parsedChannelWidth = UCentralUtils
+					.parseChannelWidth(state.radios[radioIndex].channel_width);
+				if (parsedChannelWidth != null) {
+					currentChannelWidth = parsedChannelWidth;
+					break;
 				}
-				break;
+
+				logger.error(
+					"Invalid channel width {}",
+					state.radios[radioIndex].channel_width
+				);
+				continue;
 			}
 		}
 		return new int[] { currentChannel, currentChannelWidth };
--- a/src/main/java/com/facebook/openwifirrm/optimizers/channel/ChannelOptimizer.java
+++ b/src/main/java/com/facebook/openwifirrm/optimizers/channel/ChannelOptimizer.java
@@ -156,7 +156,7 @@ public abstract class ChannelOptimizer {
 			.removeIf(serialNumber -> !deviceConfigs.containsKey(serialNumber));
 		this.model.latestState.keySet()
 			.removeIf(serialNumber -> !deviceConfigs.containsKey(serialNumber));
-		this.model.latestDeviceStatus.keySet()
+		this.model.latestDeviceStatusRadios.keySet()
 			.removeIf(serialNumber -> !deviceConfigs.containsKey(serialNumber));
 		this.model.latestDeviceCapabilities.keySet()
 			.removeIf(serialNumber -> !deviceConfigs.containsKey(serialNumber));
--- a/src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPC.java
+++ b/src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPC.java
@@ -350,20 +350,14 @@ public class MeasurementBasedApApTPC extends TPC {
 				}
 
 				for (State.Interface.SSID ssid : iface.ssids) {
-					String radioRef = ssid.radio.get("$ref").getAsString();
-					int idx = -1;
-					try {
-						idx = Integer.parseInt(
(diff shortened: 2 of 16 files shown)


src/main/java/com/facebook/openwifirrm/ucentral/models/State.java
```diff
@@ -112,8 +112,21 @@ public class Memory {
 
 	public Unit unit;
 
+	public class Radio {
+		public long active_ms;
+		public long busy_ms;
+		public int channel;
+		public String channel_width;
```
elludraon, line 119: note: looks like String is correct
- https://github.com/blogic/ucentral-schema/blob/10373083fb408ce50056b770decefd354c0f820b/renderer/wifi/iface.uc#L20-L28
- https://github.com/blogic/ucentral-schema/blob/85104eb47e25abb103b4f27e037a2392e617593c/system/state.uc#L152

src/test/java/com/facebook/openwifirrm/optimizers/TestUtils.java
```diff
@@ -397,7 +397,7 @@ private static State.Interface createDownStateInterface(int index) {
 	}
 
 	/** Create an element of {@link State#radios}. */
-	private static JsonObject createStateRadio() {
+	private static State.Radio createStateRadio() {
 		// @formatter:off
 		return gson.fromJson(
```
elludraon, line 402: While you're at it, can you replace the body of this function with a regular constructor (not gson)?

src/main/java/com/facebook/openwifirrm/ucentral/models/State.java
```diff
@@ -112,8 +112,21 @@ public class Memory {
 
 	public Unit unit;
 
+	public class Radio {
+		public long active_ms;
+		public long busy_ms;
+		public int channel;
```
elludraon, line 118: I'm not 100% sure these fields were always non-null... I think it's probably safe to assume `channel` 0 = uninitialized, less sure about `tx_power` 0 since that's a valid value(?). Let's be careful here and do some more testing... for example if you can reset an AP to default *configuration* (can you do this by deleting "channel" and "tx-power" keys?) then what do the stats (this object) return?

src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApClientTPC.java
```diff
@@ -154,19 +153,18 @@ private double computeTxPower(
...
-				!radio.get("channel_width").isJsonNull()
-					? radio.get("channel_width").getAsInt()
-					: 20);
+		int currentTxPower = radio.tx_power;
+		int channelWidth = 20;
+		try {
+			channelWidth = Integer.parseInt(radio.channel_width);
+		} catch (NumberFormatException e) {
+			logger.error("Could not parse channel width in state radio", e);
+		}
```
elludraon, line 166: move to `UCentralUtils` method, handle `80p80`

src/main/java/com/facebook/openwifirrm/optimizers/channel/ChannelOptimizer.java
```diff
@@ -375,14 +375,16 @@ protected static int[] getCurrentChannel(
...
-				currentChannelWidth = state.radios[radioIndex]
-					.get("channel_width")
-					.getAsInt();
+				try {
+					currentChannelWidth = Integer
+						.parseInt(state.radios[radioIndex].channel_width);
+				} catch (NumberFormatException e) {
+					logger.error("Invalid channel width", e);
+					continue;
+				}
```
elludraon, line 387: see other comment

elludraon (commented):
This is good, but next time separate the refactoring PR from the functionality change as it's pretty buried in here.

src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPC.java
```diff
@@ -369,38 +342,74 @@ protected void buildTxPowerMapForChannel(
...
+						);
+					} catch (NumberFormatException e) {
+						logger.error(
+							"Unable to get radio for {}, invalid radio ref",
+							e,
+							serialNumber
+						);
+						continue;
+					}
+					State.Radio radio = deviceState.radios[idx];
```
elludraon, line 370: move to `UCentralUtils` (not the logger call though, keep that here)

src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPC.java
```diff
@@ -370,40 +340,76 @@ protected void buildTxPowerMapForChannel(
...
-				// this AP is not on the band of interest
-				continue;
-			}
-			int currentTxPower = possibleCurrentTxPower.get();
-			String bssid = state.interfaces[0].ssids[0].bssid;
-			List<Integer> rssiValues = bssidToRssiValues.get(bssid);
-			logger.debug("Device <{}> : BSSID <{}>", serialNumber, bssid);
-			for (int rssi : rssiValues) {
-				logger.debug("  Neighbor received RSSI: {}", rssi);
+			State deviceState =
```
RockyMandayam2, line 343: These two lines can fit in one line

src/test/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPCTest.java
```diff
@@ -298,25 +297,6 @@ void testGetManagedBSSIDs() throws Exception {
 
 	@Test
 	@Order(2)
-	void testGetCurrentTxPower() throws Exception {
```
RockyMandayam2, line 301: Why is this test gone?

jwoos: method is gone - it doesn't make sense to have it anymore imo since the field is strongly typed now.

src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPC.java
jwoos: oops forgot to run spotless I think

src/main/java/com/facebook/openwifirrm/ucentral/models/State.java
jwoos: will try to see how I can test this more

src/main/java/com/facebook/openwifirrm/optimizers/tpc/MeasurementBasedApApTPC.java
jwoos: never mind, i did run it and didn't do anything. will change

src/main/java/com/facebook/openwifirrm/ucentral/models/State.java
jwoos: Tried deleting the configuration, if tx power and channel are not available then `radios` is just not populated at all

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
```diff
@@ -440,4 +440,43 @@ public static String getBandFromChannel(int channel) {
...
+	/**
+	 * Tries to parse the index from the reference string in the JSON returned from
+	 * other services. Note that this only returns the index, the caller is
+	 * responsible for making sure that the correct field is passed in and the
+	 * index is used in the correct fields. If there's an error parsing, it will
+	 * return null.
+	 *
+	 * @param reference The reference string, keyed under `$ref`
+	 * @return the index of the reference or null if an error occurred.
+	 * @
```
elludraon, line 468: delete

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
```diff
@@ -440,4 +440,43 @@ public static String getBandFromChannel(int channel) {
...
+	}
+
+	/**
+	 * Tries to parse the index from the reference string in the JSON returned from
+	 * other services. Note that this only returns the index, the caller is
+	 * responsible for making sure that the correct field is passed in and the
+	 * index is used in the correct fields. If there's an error parsing, it will
+	 * return null.
+	 *
+	 * @param reference The reference string, keyed under `$ref`
```
elludraon, line 466: don't use backticks in javadoc, just put in regular quotes or `{@code xx}`

elludraon (approved):

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
```diff
@@ -440,4 +440,53 @@ public static String getBandFromChannel(int channel) {
 		}
 		return null;
 	}
+
+	/**
+	 * Tries to parse channel width, if it encounters an error it will return null.
+	 * It can handle 80p80 in two ways. First it can just treat it as 160. Second,
+	 * it can just apply to the first 80 channel and ignore the second. This is
+	 * controlled by treatSeparate. For purposes of TPC algorithms, we treat it as
+	 * 160 while channel assignment treats it as separate.
```
elludraon, line 449: This last sentence shouldn't be here, avoid putting comments in places where it'll get easily forgotten/outdated

elludraon (approved):

PR #78, opened 2022-09-17
Discard unneeded ies

RockyMandayam2:
Discard unneeded ies and keep the relevant ones:
- Remove the ies field
- Parse out specific ies and put them in their own fields

There are other relevant IEs, but they won't be used just yet and their accuracy needs to be confirmed, so they will come in a separate PR.

### DIFF
--- a/src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
+++ b/src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
@@ -22,8 +22,8 @@ import com.facebook.openwifirrm.aggregators.Aggregator;
 import com.facebook.openwifirrm.aggregators.MeanAggregator;
 import com.facebook.openwifirrm.modules.Modeler.DataModel;
 import com.facebook.openwifirrm.ucentral.WifiScanEntry;
-import com.facebook.openwifirrm.ucentral.operationelement.HTOperationElement;
-import com.facebook.openwifirrm.ucentral.operationelement.VHTOperationElement;
+import com.facebook.openwifirrm.ucentral.informationelement.HTOperationElement;
+import com.facebook.openwifirrm.ucentral.informationelement.VHTOperationElement;
 
 /**
  * Modeler utilities.
--- a/src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
+++ b/src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
@@ -24,6 +24,10 @@ import org.slf4j.LoggerFactory;
 import com.facebook.openwifirrm.RRMConfig;
 import com.facebook.openwifirrm.Utils;
 import com.facebook.openwifirrm.optimizers.channel.ChannelOptimizer;
+import com.facebook.openwifirrm.ucentral.informationelement.Country;
+import com.facebook.openwifirrm.ucentral.informationelement.LocalPowerConstraint;
+import com.facebook.openwifirrm.ucentral.informationelement.QbssLoad;
+import com.facebook.openwifirrm.ucentral.informationelement.TxPwrInfo;
 import com.facebook.openwifirrm.ucentral.models.State;
 import com.google.gson.Gson;
 import com.google.gson.JsonArray;
@@ -36,6 +40,9 @@ import com.google.gson.JsonObject;
 public class UCentralUtils {
 	private static final Logger logger =
 		LoggerFactory.getLogger(UCentralUtils.class);
+
+	/** Information Element (IE) content field key */
+	private static final String IE_CONTENT_FIELD_KEY = "content";
 
 	/** The Gson instance. */
 	private static final Gson gson = new Gson();
@@ -79,13 +86,59 @@ public class UCentralUtils {
 			for (JsonElement e : scanInfo) {
 				WifiScanEntry entry = gson.fromJson(e, WifiScanEntry.class);
 				entry.unixTimeMs = timestampMs;
+				extractIEs(e, entry);
 				entries.add(entry);
-
 			}
 		} catch (Exception e) {
+			logger.debug("Exception when parsing wifiscan entries", e);
 			return null;
 		}
 		return entries;
+	}
+
+	/**
+	 * Extract desired information elements (IEs) from the wifiscan entry.
+	 * Modifies {@code entry} argument. Skips invalid IEs (IEs with missing
+	 * fields).
+	 */
+	private static void extractIEs(
+		JsonElement entryJsonElement,
+		WifiScanEntry entry
+	) {
+		JsonElement iesJsonElement =
+			entryJsonElement.getAsJsonObject().get("ies");
+		if (iesJsonElement == null) {
+			logger.debug("Wifiscan entry does not contain 'ies' field.");
+			return;
+		}
+		JsonArray ies = iesJsonElement.getAsJsonArray();
+		for (JsonElement ie : ies) {
+			JsonElement typeElement = ie.getAsJsonObject().get("type");
+			if (typeElement == null) {
+				continue;
+			}
+			JsonObject contents =
+				ie.getAsJsonObject().get(IE_CONTENT_FIELD_KEY).getAsJsonObject();
+			try {
+				switch (typeElement.getAsInt()) {
+				case Country.TYPE:
+					entry.country = Country.parse(contents);
+					break;
+				case QbssLoad.TYPE:
+					entry.qbssLoad = QbssLoad.parse(contents);
+					break;
+				case LocalPowerConstraint.TYPE:
+					entry.localPowerConstraint =
+						LocalPowerConstraint.parse(contents);
+					break;
+				case TxPwrInfo.TYPE:
+					entry.txPwrInfo = TxPwrInfo.parse(contents);
+					break;
+				}
+			} catch (NullPointerException e) {
+				logger.debug("Skipping invalid IE.", e);
+			}
+		}
 	}
 
 	/**
--- a/src/main/java/com/facebook/openwifirrm/ucentral/WifiScanEntry.java
+++ b/src/main/java/com/facebook/openwifirrm/ucentral/WifiScanEntry.java
@@ -10,6 +10,10 @@ package com.facebook.openwifirrm.ucentral;
 
 import java.util.Objects;
 
+import com.facebook.openwifirrm.ucentral.informationelement.Country;
+import com.facebook.openwifirrm.ucentral.informationelement.LocalPowerConstraint;
+import com.facebook.openwifirrm.ucentral.informationelement.QbssLoad;
+import com.facebook.openwifirrm.ucentral.informationelement.TxPwrInfo;
 import com.facebook.openwifirrm.ucentral.models.WifiScanEntryResult;
 
 /**
@@ -22,6 +26,10 @@ public class WifiScanEntry extends WifiScanEntryResult {
 	 * time reference.
 	 */
 	public long unixTimeMs;
+	public Country country;
+	public QbssLoad qbssLoad;
+	public LocalPowerConstraint localPowerConstraint;
+	public TxPwrInfo txPwrInfo;
 
 	/** Default Constructor. */
 	public WifiScanEntry() {}
@@ -30,13 +38,23 @@ public class WifiScanEntry extends WifiScanEntryResult {
 	public WifiScanEntry(WifiScanEntry o) {
 		super(o);
 		this.unixTimeMs = o.unixTimeMs;
+		this.country = o.country;
+		this.qbssLoad = o.qbssLoad;
+		this.localPowerConstraint = o.localPowerConstraint;
+		this.txPwrInfo = o.txPwrInfo;
 	}
 
 	@Override
 	public int hashCode() {
 		final int prime = 31;
 		int result = super.hashCode();
-		result = prime * result + Objects.hash(unixTimeMs);
+		result = prime * result + Objects.hash(
+			country,
+			localPowerConstraint,
+			qbssLoad,
+			txPwrInfo,
+			unixTimeMs
+		);
 		return result;
 	}
 
@@ -52,7 +70,12 @@ public class WifiScanEntry extends WifiScanEntryResult {
 			return false;
 		}
 		WifiScanEntry other = (WifiScanEntry) obj;
-		return unixTimeMs == other.unixTimeMs;
+		return Objects.equals(country, other.country) && Objects.equals(
+			localPowerConstraint,
+			other.localPowerConstraint
+		) && Objects.equals(qbssLoad, other.qbssLoad) && Objects
+			.equals(txPwrInfo, other.txPwrInfo) &&
+			unixTimeMs == other.unixTimeMs;
 	}
 
 	@Override
@@ -64,4 +87,5 @@ public class WifiScanEntry extends WifiScanEntryResult {
 			unixTimeMs
 		);
 	}
+
 }
--- a/src/main/java/com/facebook/openwifirrm/ucentral/informationelement/Country.java
+++ b/src/main/java/com/facebook/openwifirrm/ucentral/informationelement/Country.java
@@ -1,0 +1,165 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+package com.facebook.openwifirrm.ucentral.informationelement;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+
+/**
+ * This information element (IE) appears in wifiscan entries.
+ * Refer to the 802.11 specification for more details. Language in
+ * javadocs is taken from the specification.
+ */
+public class Country {
+	private static final Logger logger = LoggerFactory.getLogger(Country.class);
+
+	/** Defined in 802.11 */
+	public static final int TYPE = 7;
+
+	public static class CountryInfo {
+		/**
+		 * The lowest channel number in the CountryInfo.
+		 */
+		public final int firstChannelNumber;
+		/**
+		 * The maximum power, in dBm, allowed to be transmitted.
+		 */
+		public final int maximumTransmitPowerLevel;
+		/**
+		 * Number of channels this CountryInfo applies to. E.g., if First
+		 * Channel Number is 2 and Number of Channels is 4, this CountryInfo
+		 * describes channels 2, 3, 4, and 5.
+		 */
+		public final int numberOfChannels;
+
+		/** Constructor. */
+		public CountryInfo(
+			int firstChannelNumber,
+			int maximumTransmitPowerLevel,
+			int numberOfChannels
+		) {
+			this.firstChannelNumber = firstChannelNumber;
+			this.maximumTransmitPowerLevel = maximumTransmitPowerLevel;
+			this.numberOfChannels = numberOfChannels;
+		}
+
+		/** Parse CountryInfo from the appropriate Json object. */
+		public static CountryInfo parse(JsonObject contents) {
+			final int firstChannelNumber =
+				contents.get("First Channel Number").getAsInt();
+			final int maximumTransmitPowerLevel = contents
+				.get("Maximum Transmit Power Level (in dBm)")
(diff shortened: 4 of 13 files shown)


src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -22,8 +22,8 @@
 import com.facebook.openwifirrm.aggregators.MeanAggregator;
 import com.facebook.openwifirrm.modules.Modeler.DataModel;
 import com.facebook.openwifirrm.ucentral.WifiScanEntry;
-import com.facebook.openwifirrm.ucentral.operationelement.HTOperationElement;
-import com.facebook.openwifirrm.ucentral.operationelement.VHTOperationElement;
+import com.facebook.openwifirrm.ucentral.informationelement.HTOperation;
```
elludraon, line 25: Please don't include these refactoring changes along with actual changes

src/main/java/com/facebook/openwifirrm/ucentral/models/WifiScanEntryResult.java
```diff
@@ -50,8 +53,10 @@ public class WifiScanEntryResult {
 	public String vht_oper;
 	public int capability;
 	public int frequency;
-	/** IE = information element */
-	public JsonArray ies;
+	public Country country;
+	public QbssLoad qbssLoad;
+	public LocalPowerConstraint localPowerConstraint;
+	public TxPwrInfo txPwrInfo;
```
elludraon, line 59: This should go in the subclass, don't add it to the ucentral model

src/main/java/com/facebook/openwifirrm/ucentral/models/WifiScanEntryResult.java
```diff
@@ -50,8 +48,6 @@ public class WifiScanEntryResult {
 	public String vht_oper;
 	public int capability;
 	public int frequency;
-	/** IE = information element */
-	public JsonArray ies;
```
elludraon, line 54: can you put a comment here like:
```
// ies[] array is not stored directly, but parsed into WifiScanEntry fields
```

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
```diff
@@ -79,10 +83,45 @@ public static List<WifiScanEntry> parseWifiScanEntries(
 			for (JsonElement e : scanInfo) {
 				WifiScanEntry entry = gson.fromJson(e, WifiScanEntry.class);
 				entry.unixTimeMs = timestampMs;
+				// extract relevant IEs
```
elludraon, line 86: can you move this into a separate method? (this function is getting too long, and for example we may want to unit test IE parsing only)

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
```diff
@@ -79,10 +83,45 @@ public static List<WifiScanEntry> parseWifiScanEntries(
...
+				JsonArray ies = e.getAsJsonObject().get("ies").getAsJsonArray();
```
elludraon, line 87: do error handling here (don't error out if "ies" isn't present, for example)

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
```diff
@@ -79,10 +83,45 @@ public static List<WifiScanEntry> parseWifiScanEntries(
...
+					switch (typeElement.getAsInt()) {
+					case Country.TYPE:
+						entry.country = gson.fromJson(
+							ie.getAsJsonObject().get(contentFieldKey),
+							Country.class
+						);
+						break;
+					case QbssLoad.TYPE:
+						entry.qbssLoad = gson.fromJson(
+							ie.getAsJsonObject().get(contentFieldKey).getAsJsonObject().get("802.11e CCA Version"),
```
elludraon, line 103: can you move the parsing of the inner "802.11e CCA Version" part into the `QbssLoad` class as well? otherwise it feels like you're leaking implementation details into this parser function

src/main/java/com/facebook/openwifirrm/ucentral/WifiScanEntry.java
```diff
@@ -22,6 +26,10 @@ public class WifiScanEntry extends WifiScanEntryResult {
 	 * time reference.
 	 */
 	public long unixTimeMs;
+	public Country country;
+	public QbssLoad qbssLoad;
+	public LocalPowerConstraint localPowerConstraint;
+	public TxPwrInfo txPwrInfo;
```
elludraon, line 32: I think I would prefer seeing a separate class/field like `InformationElement ie` to hold all of these as sub-fields, do you agree?

Also, do we actually need equality checks on `WifiScanEntry`? If not then I wouldn't bother putting any of those autogenerated overrides since it's adding a lot of bloat to these files...

src/main/java/com/facebook/openwifirrm/ucentral/informationelement/Country.java
```diff
@@ -0,0 +1,173 @@
...
+	public Country(List<CountryInfo> countryInfos) {
+		this.constraints = Collections.unmodifiableList(countryInfos);
+	}
+
+	/**
+	 * ConstraintsDeserializer is used because the constraints field in the
+	 * json is "nested" (instead of a list of CountryInfos, it is a list of
+	 * JsonObjects each with exactly one key called "Country Info").
+	 */
+	private static class ConstraintsDeserializer
```
elludraon, line 126: ...I am not sure how I feel about using gson to construct these IE classes, and seeing these custom deserializer classes especially makes me lean towards "no". The alternative is only *slightly* more code (ex. static method like `Country.parse(JsonObject contents) -> Country/null`) since manual parsing is so straightforward, and that might even end up being less code than a single custom deserializer class (and far easier to read/understand).

This is assuming we never need to re-serialize which is probably true.

src/main/java/com/facebook/openwifirrm/ucentral/WifiScanEntry.java
RockyMandayam2: I don't have much of a preference regarding an `InformationElement` class. My thoughts/questions:
- Note that these classes are already part of one package called `informationelement`
- Would there be any methods in `InformationElement`?
- Probably not a big deal, but the `ht_oper` and `vht_oper` fields in `WifiScanEntryResult` are their separate fields (both in the parsed json and in our data structure), but really, they are also IEs like any other according 802.11. We could not include these two in `InformationElement` since they are in the uCentral defined `WifiscanEntryResult`.

Regarding equality checks, we use them for unit tests. But they may not need to include every field...

RockyMandayam2: Also, if we do the new class, maybe it should be `InformationElements` since it is containing multiple IEs, and is not itself an IE?

src/main/java/com/facebook/openwifirrm/ucentral/informationelement/Country.java
RockyMandayam2: Ok, I will try a `static parse(JsonObject contents)` for every IE (except `HTOperationElement` and `VHTOperationElement` which are parsed and stored differently).

src/main/java/com/facebook/openwifirrm/ucentral/WifiScanEntry.java
```diff
@@ -30,13 +38,23 @@ public WifiScanEntry() {}
...
+		this.localPowerConstraint = o.localPowerConstraint;
+		this.txPwrInfo = o.txPwrInfo;
 	}
 
 	@Override
 	public int hashCode() {
 		final int prime = 31;
 		int result = super.hashCode();
-		result = prime * result + Objects.hash(unixTimeMs);
+		result = prime * result + Objects.hash(
```
wuweicaifb, line 51: I think we should use `<<` instead of `*`.

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
```diff
@@ -79,15 +83,62 @@ public static List<WifiScanEntry> parseWifiScanEntries(
...
+			logger.debug("Wifiscan entry does not contain 'ies' field.");
+			return;
+		}
+		JsonArray ies = iesJsonElement.getAsJsonArray();
+		for (JsonElement ie : ies) {
+			JsonElement typeElement = ie.getAsJsonObject().get("type");
+			if (typeElement == null) {
+				continue;
+			}
+			final String contentFieldKey = "content";
```
wuweicaifb, line 117: I'd prefer to make it a static final for the class, but feel free to push me back.

src/main/java/com/facebook/openwifirrm/ucentral/informationelement/Country.java
```diff
@@ -0,0 +1,171 @@
...
+		 */
+		public final int firstChannelNumber;
+		/**
+		 * The maximum power, in dBm, allowed to be transmitted.
+		 */
+		public final int maximumTransmitPowerLevel;
+		/**
+		 * Number of channels this CountryInfo applies to. E.g., if First
+		 * Channel Number is 2 and Number of Channels is 4, this CountryInfo
+		 * describes channels 2, 3, 4, and 5.
```
wuweicaifb, line 45: I don't think channels have to be consecutive numbers

src/main/java/com/facebook/openwifirrm/ucentral/informationelement/Country.java
```diff
@@ -0,0 +1,171 @@
...
+	public final List<CountryInfo> constraints;
+
+	/**
+	 * Constructor. This constructor is currently never used, but it is
+	 * required by Java since at least one member variable is marked final.
+	 */
+	public Country(List<CountryInfo> countryInfos) {
+		this.constraints = Collections.unmodifiableList(countryInfos);
+	}
+
```
wuweicaifb, line 126: Maybe I'm wrong, but it looks weird to me that an object can only be constructed by a list of objects of its internal class. Feel free to push me back

wuweicaifb (changes requested):

src/main/java/com/facebook/openwifirrm/ucentral/models/WifiScanEntryResult.java
```diff
@@ -50,8 +48,7 @@ public class WifiScanEntryResult {
...
+	// ies[] array is not stored directly, but parsed into WifiScanEntry fields
```
wuweicaifb, line 51: Just remove the comment here? Or move it to the javadoc of this class.

src/main/java/com/facebook/openwifirrm/ucentral/informationelement/TxPwrInfo.java
```diff
@@ -0,0 +1,118 @@
...
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+
+/**
+ * This information element (IE) appears in wifiscan entries. It is called
+ * "Tx Pwr Info" in these entries, and "Transmit Power Envelope" in the 802.11
+ * specification. Refer to the specification for more details. Language in
+ * javadocs is taken from the specification.
+ */
+public class TxPwrInfo {
```
wuweicaifb, line 22: Is `TxPwrInfo` used for consistency with the wifi scan json? `txPower` is used in more places in the codebase, so I wonder if we can keep them consistent among the codebase.

src/main/java/com/facebook/openwifirrm/ucentral/models/WifiScanEntryResult.java
RockyMandayam2: Jeffrey wanted the comment. But I can move it to javadoc

src/main/java/com/facebook/openwifirrm/ucentral/informationelement/TxPwrInfo.java
RockyMandayam2: `txPower` is usually used as a method parameter or local variable or something. This is specifically for the Information Element and only describes constraints (max tx power limits)

src/main/java/com/facebook/openwifirrm/ucentral/informationelement/Country.java
RockyMandayam2: Hm, good point... I don't know of a better way. At least, since the nested class is static and public, I don't think it's a big deal. Anyone else have thoughts?

src/main/java/com/facebook/openwifirrm/ucentral/informationelement/Country.java
RockyMandayam2: Correct, but this is just how the standard describes the limits on the channel. At least, that's what I understand from the "9.4.2.8 Country element" section.

src/main/java/com/facebook/openwifirrm/ucentral/informationelement/Country.java
wuweicaifb: I'm fine with the current way, but just wondering if there's a better way. Open to any idea here.

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
RockyMandayam2: I'll make it `private static final`

src/main/java/com/facebook/openwifirrm/ucentral/WifiScanEntry.java
wuweicaifb: Based on offline discussion, I made a mistake here. Please ignore the request.

src/main/java/com/facebook/openwifirrm/ucentral/WifiScanEntry.java
wuweicaifb: At the very beginning, I also preferred a class like `InformationElements` to include new fields so that it's clear that which ones are ies, but now I think your opinion 3 is strong enough to leave them separate without a umbrella class to hold them.

wuweicaifb (approved):

src/main/java/com/facebook/openwifirrm/ucentral/informationelement/LocalPowerConstraint.java
```diff
@@ -0,0 +1,75 @@
...
+ * Language in javadocs is taken from the specification.
+ */
+public class LocalPowerConstraint {
+
+	/** Defined in 802.11 */
+	public static final int TYPE = 32;
+
+	/**
+	 * Units are dB.
+	 * <p>
```
zhiqiand, line 28: Why there is `<p>`?

src/main/java/com/facebook/openwifirrm/ucentral/informationelement/QbssLoad.java
```diff
@@ -0,0 +1,114 @@
...
+
+	/** Parse QbssLoad IE from appropriate Json object. */
+	public static QbssLoad parse(JsonObject contents) {
+		contents = contents.get("802.11e CCA Version").getAsJsonObject();
+		final int stationCount = contents.get("Station Count").getAsInt();
+		final int channelUtilization =
+			contents.get("Channel Utilization").getAsInt();
+		final int availableAdmissionCapacity =
+			contents.get("Available Admission Capabilities").getAsInt();
+		return new QbssLoad(
```
zhiqiand, line 75: Actually the constructor is used here?

zhiqiand (commented):
Leave some minor comments. Besides, does the equality checks used? I see that `equals` is override but didn't see if it used some where.

src/main/java/com/facebook/openwifirrm/ucentral/WifiScanEntry.java
RockyMandayam2: @zhiqiand since you mentioned equality checks, here is a thread that discusses it (and another concern). What are your thoughts?

src/main/java/com/facebook/openwifirrm/ucentral/informationelement/LocalPowerConstraint.java
RockyMandayam2: It is a javadoc formatting thing. It adds a new paragraph.

src/main/java/com/facebook/openwifirrm/ucentral/informationelement/QbssLoad.java
RockyMandayam2: Good catch, I will update all constructor javadocs

src/main/java/com/facebook/openwifirrm/ucentral/informationelement/Country.java
RockyMandayam2: Zhiqian also mentioned it may be fine since the inner class is static

src/main/java/com/facebook/openwifirrm/ucentral/WifiScanEntry.java
zhiqiand: I just want to make sure if the equality checks are used somewhere or I missed it. The equality check I mentioned is the classes under `informationelement` folder. If it is not used then it's better not to add it at this moment.

RockyMandayam2: I wanted equality checks for `WifiScanEntry` (because equality checks for that class are used in unit testing). Since these classes are part of `WifiScanEntry` they should also implement equality checks.

However, since these specific classes aren't actually what is tested in the unit tests, I could remove them and it wouldn't cause any problems now (though if we add individual IE testing later, it may cause problems).

Thoughts?

jwoos (approved):

src/main/java/com/facebook/openwifirrm/ucentral/informationelement/Country.java
```diff
@@ -0,0 +1,165 @@
...
+ * Refer to the 802.11 specification for more details. Language in
+ * javadocs is taken from the specification.
+ */
+public class Country {
+	private static final Logger logger = LoggerFactory.getLogger(Country.class);
+
+	/** Defined in 802.11 */
+	public static final int TYPE = 7;
+
+	public static class CountryInfo {
```
elludraon, line 33: missing javadoc (all public elements need docs)

src/main/java/com/facebook/openwifirrm/ucentral/informationelement/Country.java
```diff
@@ -0,0 +1,165 @@
...
+	/** Parse Country IE from the appropriate Json object. */
+	public static Country parse(JsonObject contents) {
+		List<CountryInfo> constraints = new ArrayList<>();
+		JsonElement constraintsObject = contents.get("constraints");
+		if (constraintsObject != null) {
+			for (JsonElement jsonElement : constraintsObject.getAsJsonArray()) {
+				try {
+					constraints
+						.add(CountryInfo.parse(jsonElement.getAsJsonObject()));
+				} catch (NullPointerException e) {
```
elludraon, line 130: same here

src/main/java/com/facebook/openwifirrm/ucentral/WifiScanEntry.java
RockyMandayam2: Will address in a new PR

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
```diff
@@ -79,15 +86,61 @@ public static List<WifiScanEntry> parseWifiScanEntries(
...
+					break;
+				case LocalPowerConstraint.TYPE:
+					entry.localPowerConstraint =
+						LocalPowerConstraint.parse(contents);
+					break;
+				case TxPwrInfo.TYPE:
+					entry.txPwrInfo = TxPwrInfo.parse(contents);
+					break;
+				}
+			} catch (NullPointerException e) {
```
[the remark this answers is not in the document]
RockyMandayam2: Will address in separate PR

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
```diff
@@ -79,15 +86,61 @@ public static List<WifiScanEntry> parseWifiScanEntries(
...
+				logger.debug("Skipping invalid IE.", e);
```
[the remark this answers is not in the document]
RockyMandayam2: Good idea, good to know. (I'm removing this block though, so won't apply here)

src/main/java/com/facebook/openwifirrm/ucentral/informationelement/Country.java
RockyMandayam2: Will address in separate PR

PR #80, opened 2022-09-20
Only push updates for desired zones/venues

RockyMandayam2:
Only push updates for desired zones/venues

### DIFF
--- a/src/main/java/com/facebook/openwifirrm/RRMAlgorithm.java
+++ b/src/main/java/com/facebook/openwifirrm/RRMAlgorithm.java
@@ -143,6 +143,8 @@ public class RRMAlgorithm {
 	 * @param dryRun if set, do not apply changes
 	 * @param allowDefaultMode if false, "mode" argument must be present and
 	 *                         valid (returns error if invalid)
+	 * @param updateImmediately whether to interrupt the config manager thread
+	 * to trigger immediate update
 	 *
 	 * @return the algorithm result, with exactly one field set ("error" upon
 	 *         failure, any others upon success)
@@ -153,7 +155,8 @@ public class RRMAlgorithm {
 		Modeler modeler,
 		String zone,
 		boolean dryRun,
-		boolean allowDefaultMode
+		boolean allowDefaultMode,
+		boolean updateImmediately
 	) {
 		AlgorithmResult result = new AlgorithmResult();
 		if (name == null || args == null) {
@@ -212,11 +215,14 @@ public class RRMAlgorithm {
 			}
 			result.channelMap = optimizer.computeChannelMap();
 			if (!dryRun) {
-				optimizer.applyConfig(
+				optimizer.updateDeviceApConfig(
 					deviceDataManager,
 					configManager,
 					result.channelMap
 				);
+				if (updateImmediately) {
+					configManager.queueZoneAndWakeUp(zone);
+				}
 			}
 		} else if (
 			name.equals(RRMAlgorithm.AlgorithmType.OptimizeTxPower.name())
@@ -270,16 +276,18 @@ public class RRMAlgorithm {
 			}
 			result.txPowerMap = optimizer.computeTxPowerMap();
 			if (!dryRun) {
-				optimizer.applyConfig(
+				optimizer.updateDeviceApConfig(
 					deviceDataManager,
 					configManager,
 					result.txPowerMap
 				);
+				if (updateImmediately) {
+					configManager.queueZoneAndWakeUp(zone);
+				}
 			}
 		} else {
 			result.error = String.format("Unknown algorithm: '%s'", name);
 		}
-		configManager.wakeUp();
 		return result;
 	}
 }
--- a/src/main/java/com/facebook/openwifirrm/RRMAlgorithm.java
+++ b/src/main/java/com/facebook/openwifirrm/RRMAlgorithm.java
@@ -143,8 +143,9 @@ public class RRMAlgorithm {
 	 * @param dryRun if set, do not apply changes
 	 * @param allowDefaultMode if false, "mode" argument must be present and
 	 *                         valid (returns error if invalid)
-	 * @param updateImmediately whether to interrupt the config manager thread
-	 * to trigger immediate update
+	 * @param updateImmediately true if the method should queue the zone for
+	 * 							update and interrupt the config manager thread
+	 * 							to trigger immediate update
 	 *
 	 * @return the algorithm result, with exactly one field set ("error" upon
 	 *         failure, any others upon success)
--- a/src/main/java/com/facebook/openwifirrm/RRMAlgorithm.java
+++ b/src/main/java/com/facebook/openwifirrm/RRMAlgorithm.java
@@ -279,6 +279,7 @@ public class RRMAlgorithm {
 		} else {
 			result.error = String.format("Unknown algorithm: '%s'", name);
 		}
+		configManager.wakeUp();
 		return result;
 	}
 }
--- a/src/main/java/com/facebook/openwifirrm/modules/ApiServer.java
+++ b/src/main/java/com/facebook/openwifirrm/modules/ApiServer.java
@@ -917,7 +917,7 @@ public class ApiServer implements Runnable {
 				DeviceConfig networkConfig =
 					gson.fromJson(request.body(), DeviceConfig.class);
 				deviceDataManager.setDeviceNetworkConfig(networkConfig);
-				configManager.wakeUp();
+				configManager.wakeUp(null);
 
 				// Revalidate data model
 				modeler.revalidate();
@@ -981,7 +981,7 @@ public class ApiServer implements Runnable {
 				DeviceConfig zoneConfig =
 					gson.fromJson(request.body(), DeviceConfig.class);
 				deviceDataManager.setDeviceZoneConfig(zone, zoneConfig);
-				configManager.wakeUp();
+				configManager.wakeUp(zone);
 
 				// Revalidate data model
 				modeler.revalidate();
@@ -1044,7 +1044,9 @@ public class ApiServer implements Runnable {
 				DeviceConfig apConfig =
 					gson.fromJson(request.body(), DeviceConfig.class);
 				deviceDataManager.setDeviceApConfig(serialNumber, apConfig);
-				configManager.wakeUp();
+				final String zone =
+					deviceDataManager.getDeviceZone(serialNumber);
+				configManager.wakeUp(zone);
 
 				// Revalidate data model
 				modeler.revalidate();
@@ -1117,7 +1119,9 @@ public class ApiServer implements Runnable {
 						.computeIfAbsent(serialNumber, k -> new DeviceConfig())
 						.apply(apConfig);
 				});
-				configManager.wakeUp();
+				final String zone =
+					deviceDataManager.getDeviceZone(serialNumber);
+				configManager.wakeUp(zone);
 
 				// Revalidate data model
 				modeler.revalidate();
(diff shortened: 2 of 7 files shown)


src/main/java/com/facebook/openwifirrm/modules/ApiServer.java
```diff
@@ -1117,7 +1119,9 @@ public String handle(
 						.computeIfAbsent(serialNumber, k -> new DeviceConfig())
 						.apply(apConfig);
 				});
-				configManager.wakeUp();
+				final String zone =
+					deviceDataManager.getDeviceZone(serialNumber);
+				configManager.wakeUp(zone);
```
elludraon, line 1124: Hm.... perhaps we want to actually have a separate call for specific devices. Put a TODO here for now since it doesn't make sense to trigger zone update for a single AP change.

src/main/java/com/facebook/openwifirrm/modules/ConfigManager.java
```diff
@@ -63,8 +65,14 @@ private class DeviceData {
 	/** Is the main thread sleeping? */
 	private final AtomicBoolean sleepingFlag = new AtomicBoolean(false);
 
-	/** Was a manual config update requested? */
-	private final AtomicBoolean eventFlag = new AtomicBoolean(false);
+	/**
+	 * Set of venues for which manual config updates have been requested.
+	 * This is a thread-safe (concurrent, not synchronized) set, since it is
```
elludraon, line 70: what does "concurrent, not synchronized" mean??

src/main/java/com/facebook/openwifirrm/modules/ConfigManager.java
```diff
@@ -63,8 +65,14 @@ private class DeviceData {
...
+	 * backed by a {@code ConcurrentHashMap}, but is still statically typed as
+	 * {@code Set<String>} since there is nothing like ConcurrentHashSet in
+	 * Java.
```
elludraon, line 73: you do not need to explain what `ConcurrentHashMap` is in this comment, readers can look it up themselves

src/main/java/com/facebook/openwifirrm/modules/ConfigManager.java
```diff
@@ -364,9 +376,23 @@ public boolean removeConfigListener(String id) {
 		return (configListeners.remove(id) != null);
 	}
 
-	/** Interrupt the main thread, possibly triggering an update immediately. */
-	public void wakeUp() {
-		eventFlag.set(true);
+	/**
+	 * Interrupt the main thread, possibly triggering an update immediately.
+	 *
+	 * @param venue venue, or null, which indicates all venues
```
elludraon, line 382: our terminology is "zone" ("venue" is the Prov term)

src/main/java/com/facebook/openwifirrm/modules/ConfigManager.java
```diff
@@ -364,9 +376,23 @@ public boolean removeConfigListener(String id) {
...
+	 */
+	public void wakeUp(String venue) {
+		if (venue != null) {
+			venuesToUpdate.add(venue);
+		} else {
+			/*
+			 * Here, addAll is not atomic, so read operations during the addAll
+			 * may none, some, or all of the items passed in to addAll. But, it
+			 * does not matter, since no thread reads venuesToUpdate.
```
elludraon, line 391: the main thread is reading this (?)

src/main/java/com/facebook/openwifirrm/modules/ConfigManager.java
RockyMandayam2: I'm not exactly using ConcurrentHashMap (I'm calling `newKeySet()` which returns a thread-safe set). Do you think I should remove most of this javadoc?

src/main/java/com/facebook/openwifirrm/modules/ConfigManager.java
RockyMandayam2: I will add a quick explanation

src/main/java/com/facebook/openwifirrm/modules/ConfigManager.java
```diff
@@ -63,8 +66,27 @@ private class DeviceData {
...
+	private Set<String> zonesToUpdate = ConcurrentHashMap.newKeySet();
+
+	/**
+	 * This set is used as a buffer to track queued updates without actually
+	 * writing them to {@link #zonesToUpdate}. This is done in order to prevent
+	 * the possibility of trying to update zones in quick succession (which is
+	 * not allowed by the debounce timer).
+	 */
+	private Set<String> temporaryZonesToUpdateBuffer =
+		ConcurrentHashMap.newKeySet();
```
elludraon, line 89: I don't see the purpose of this - in what case does this actually help vs. using `zonesToUpdate` directly?

src/main/java/com/facebook/openwifirrm/modules/ConfigManager.java
```diff
@@ -256,10 +284,10 @@ private void runImpl() {
 			logger.trace("Config changes are disabled.");
 		} else if (devicesNeedingUpdate.isEmpty()) {
 			logger.debug("No device configs to send.");
-		} else if (params.configOnEventOnly && !isEvent) {
+		} else if (params.configOnEventOnly && !shouldUpdate) {
 			// shouldn't happen
 			logger.error(
-				"ERROR!! {} device(s) queued for config update, but event flag not set",
+				"ERROR!! {} device(s) queued for config update, but set of zones to update is empty.",
```
elludraon, line 290: "set of zones to update is empty" -> "no zones queued for update" (try to keep log messages plain english)

src/main/java/com/facebook/openwifirrm/modules/ConfigManager.java
```diff
@@ -201,11 +228,12 @@ private void runImpl() {
 			for (ConfigListener listener : configListeners.values()) {
 				listener.receiveDeviceConfig(device.serialNumber, data.config);
 			}
-
-			// Check event flag
+			// Check if there are requested updates for this zone
+			// And if so, remove this zone from the set of to-be-updated zones
+			boolean isEvent = zonesToUpdateCopy.remove(device.venue);
```
elludraon, line 233: - I don't understand why you are removing the zone here?? What about the rest of the devices in the zone?
- You cannot use `device.venue`, you need to look up the zone from `deviceDataManager.getDeviceZone(device.serialNumber)`.

src/main/java/com/facebook/openwifirrm/modules/ConfigManager.java
```diff
@@ -364,12 +392,44 @@ public boolean removeConfigListener(String id) {
...
+		}
+	}
+
+	/**
+	 * Interrupt the main thread, possibly triggering an update immediately.
+	 */
 	public void wakeUp() {
-		eventFlag.set(true);
+		zonesToUpdate.addAll(temporaryZonesToUpdateBuffer);
+		temporaryZonesToUpdateBuffer.clear();
```
elludraon, line 418: as mentioned above, totally don't see the purpose of the second queue or this code here

src/main/java/com/facebook/openwifirrm/modules/ApiServer.java
```diff
@@ -1044,7 +1044,9 @@ public String handle(
 				DeviceConfig apConfig =
 					gson.fromJson(request.body(), DeviceConfig.class);
 				deviceDataManager.setDeviceApConfig(serialNumber, apConfig);
-				configManager.wakeUp();
+				final String zone =
+					deviceDataManager.getDeviceZone(serialNumber);
+				configManager.wakeUp(zone);
```
elludraon, line 1049: also needs same TODO comment as below

elludraon (changes requested):
This looks mostly good (surprisingly little code, nice job!) but a few things are wrong.

src/main/java/com/facebook/openwifirrm/ucentral/gw/models/DeviceWithStatus.java
```diff
@@ -15,6 +15,7 @@
 public class DeviceWithStatus {
 	public String owner;
 	public String location;
+	/** Unreliable */
```
elludraon, line 18: ??

src/main/java/com/facebook/openwifirrm/RRMAlgorithm.java
```diff
@@ -143,6 +143,8 @@ public static RRMAlgorithm parse(String name, String argsRaw) {
 	 * @param dryRun if set, do not apply changes
 	 * @param allowDefaultMode if false, "mode" argument must be present and
 	 *                         valid (returns error if invalid)
+	 * @param updateImmediately whether to interrupt the config manager thread
+	 * to trigger immediate update
```
elludraon, line 147: Also mention that the zone gets marked for update (that's important)
style: we usually indent `@param` overflow lines, see the param above (I don't really care, just be consistent)

src/main/java/com/facebook/openwifirrm/modules/ConfigManager.java
```diff
@@ -364,9 +376,47 @@ public boolean removeConfigListener(String id) {
...
-		eventFlag.set(true);
+	/**
+	 * Mark the zone to be updated, then interrupt the main thread to possibly
+	 * trigger an update immediately.
+	 *
+	 * @param zone non-null zone (i.e., venue)
+	 */
+	public void queueZoneAndWakeUp(String zone) {
+		if (zone == null) {
+			logger.debug("Zone to queue must be a non-null String.");
```
elludraon, line 387: throw `NullPointerException` here (or at least return)?

src/main/java/com/facebook/openwifirrm/modules/ConfigManager.java
```diff
@@ -364,9 +376,47 @@ public boolean removeConfigListener(String id) {
...
+		 * possible that: only some zones are added before the main thread
+		 * enters its update loop, completes it, and goes to sleep. Then, a zone
+		 * that was initially in zonesToUpdate (but not anymore) may be added
+		 * again (as part of the addAll(...)). Then, an interrupt would trigger
+		 * an update with the latest configs for this zone, but the update would
+		 * not work due to the debounce timer. Therefore, we make it a point to
+		 * copy the state of zonesToUpdate and only add zones that are not
+		 * already in zonesToUpdate at that time.
+		 */
+		Set<String> copy = new HashSet<>(zonesToUpdate);
```
elludraon, line 409: Am I misreading this, or does it not actually do anything useful? If we hit the debounce timer then config doesn't get sent (otherwise harmless), but you are just avoiding the logic altogether (and config still doesn't get sent).

elludraon (commented):
This looks good to me! Few comments left.

src/main/java/com/facebook/openwifirrm/ucentral/gw/models/DeviceWithStatus.java
RockyMandayam2: In an earlier review, you said not to use `device.venue` in `ConfigManager` and instead to use `deviceDataManager.getDeviceZone(device.serialNumber)`

What was the reason for that?

RockyMandayam2: We discussed. It was because we may not always be syncing with prov (it may not even be deployed)

src/main/java/com/facebook/openwifirrm/modules/RRMScheduler.java
```diff
@@ -332,8 +332,10 @@ protected void performRRM(String zone) {
 				modeler,
 				zone,
 				params.dryRun,
-				true /* allowDefaultMode */
+				true, /* allowDefaultMode */
+				false /* updateImmediately */
 			);
+			configManager.queueZoneAndWakeUp(zone);
```
jwoos, line 338: do we want to do it within the loop? seems better to queue it once and wake it up after all the algos have run, especially since put `updateImmediately` as false.

jwoos (approved):
One small comment but otherwise looks good

PR #81, opened 2022-09-20
Use short instead of byte to store unsigned byte values in VHTOperationElement

RockyMandayam2:
Use `short` instead of `byte` to store unsigned byte values in VHTOperationElement. Confirmed by reading channel 155 from a vht IE from a real AP.

Will be cherry-picked into 2.7 release after this is merged into main.


elludraon (commented):
Can you use `short` instead?

elludraon (approved):

pohanhf (approved):
Looks good to me. Thanks for the fix. :)

PR #82, opened 2022-09-20
States aggregation 

zhiqiand:
### Summary
This PR aggregates State by bssid&station String and Radio. If two State matches in bssid, station, channel, channel_width and tx_power. Then the two need to be aggregated to an `AggregatedState`. In `AggregatedState`, rssi and mcs list all the values over the time.

The return aggregated map is a map from serial number to a map from bssid&station String to a list of `AggregatedState`.
`Map(Serial Number -> Map(bssid&station string -> List<AggregatedState>))`

Also keep only `latestStates` under `DataModel` instead of `latesState`. This PR fixes some bugs causing by this change.

### Implementation
For easy review, focus on `AggregatedState.java`, `ModelerUtils.java`, `ModelerUtilsTest.java` and `UtilsTest.java`.

- Create `AggregatedState` class, `AggregatedRate` and `Radio` are inner classes of `AggregatedState`. Override hashCode() under `AggregatedState` for checking if two `AggregatedStates` matches the aggregation rule. Method `AggregatedState.add()` is for easily aggregating two States of the same hashCode. 

- Mainly functional method is `ModelerUtils.getAggregatedStates()`, which aggregates all the non-obsolete State. `ModelerUtils.addStateToAggregation` is serves as a helper which aggregate one State to a list of AggregatedState.

### Follow up
- Will convert the aggregation result to json.

### DIFF
--- a/src/main/java/com/facebook/openwifirrm/modules/Modeler.java
+++ b/src/main/java/com/facebook/openwifirrm/modules/Modeler.java
@@ -8,7 +8,6 @@
 
 package com.facebook.openwifirrm.modules;
 
-import java.util.ArrayList;
 import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
@@ -272,7 +271,7 @@ public class Modeler implements Runnable {
 					State stateModel = gson.fromJson(state, State.class);
 					dataModel.latestStates.computeIfAbsent(
 						device.serialNumber,
-						k -> Collections.synchronizedList(new ArrayList<>())
+						k -> Collections.synchronizedList(new LinkedList<>())
 					).add(stateModel);
 					logger.debug(
 						"Device {}: added initial state from uCentralGw",
@@ -309,7 +308,7 @@ public class Modeler implements Runnable {
 							.computeIfAbsent(
 								record.serialNumber,
 								k -> Collections
-									.synchronizedList(new ArrayList<>())
+									.synchronizedList(new LinkedList<>())
 							);
 						if (latestStatesList.size() >= params.stateBufferSize) {
 							latestStatesList.remove(0);
--- a/src/main/java/com/facebook/openwifirrm/modules/Modeler.java
+++ b/src/main/java/com/facebook/openwifirrm/modules/Modeler.java
@@ -305,13 +305,18 @@ public class Modeler implements Runnable {
 				if (state != null) {
 					try {
 						State stateModel = gson.fromJson(state, State.class);
-						dataModel.latestStates
+						List<State> latestStatesList = dataModel.latestStates
 							.computeIfAbsent(
 								record.serialNumber,
 								k -> Collections
 									.synchronizedList(new ArrayList<>())
-							)
-							.add(stateModel);
+							);
+						if (latestStatesList.size() >= params.stateBufferSize) {
+							latestStatesList.remove(0);
+						}
+						latestStatesList.add(stateModel);
+						dataModel.latestStates
+							.put(record.serialNumber, latestStatesList);
 						stateUpdates.add(record.serialNumber);
 					} catch (JsonSyntaxException e) {
 						logger.error(
--- a/src/main/java/com/facebook/openwifirrm/modules/Modeler.java
+++ b/src/main/java/com/facebook/openwifirrm/modules/Modeler.java
@@ -9,6 +9,7 @@
 package com.facebook.openwifirrm.modules;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
@@ -271,7 +272,7 @@ public class Modeler implements Runnable {
 					State stateModel = gson.fromJson(state, State.class);
 					dataModel.latestStates.computeIfAbsent(
 						device.serialNumber,
-						k -> new ArrayList<>()
+						k -> Collections.synchronizedList(new ArrayList<>())
 					).add(stateModel);
 					logger.debug(
 						"Device {}: added initial state from uCentralGw",
@@ -307,7 +308,8 @@ public class Modeler implements Runnable {
 						dataModel.latestStates
 							.computeIfAbsent(
 								record.serialNumber,
-								k -> new ArrayList<>()
+								k -> Collections
+									.synchronizedList(new ArrayList<>())
 							)
 							.add(stateModel);
 						stateUpdates.add(record.serialNumber);
--- a/src/main/java/com/facebook/openwifirrm/modules/Modeler.java
+++ b/src/main/java/com/facebook/openwifirrm/modules/Modeler.java
@@ -8,6 +8,8 @@
 
 package com.facebook.openwifirrm.modules;
 
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
@@ -93,8 +95,9 @@ public class Modeler implements Runnable {
 		public Map<String, List<List<WifiScanEntry>>> latestWifiScans =
 			new ConcurrentHashMap<>();
 
-		/** List of latest state per device. */
-		public Map<String, State> latestState = new ConcurrentHashMap<>();
+		/** List of latest states per device. */
+		public Map<String, List<State>> latestStates =
+			new ConcurrentHashMap<>();
 
 		/** List of radio info per device. */
 		public Map<String, JsonArray> latestDeviceStatusRadios =
@@ -267,7 +270,10 @@ public class Modeler implements Runnable {
 			if (state != null) {
 				try {
 					State stateModel = gson.fromJson(state, State.class);
-					dataModel.latestState.put(device.serialNumber, stateModel);
+					dataModel.latestStates.put(
+						device.serialNumber,
+						new ArrayList<>(Arrays.asList(stateModel))
+					);
 					logger.debug(
 						"Device {}: added initial state from uCentralGw",
 						device.serialNumber
@@ -299,8 +305,12 @@ public class Modeler implements Runnable {
 				if (state != null) {
 					try {
 						State stateModel = gson.fromJson(state, State.class);
-						dataModel.latestState
-							.put(record.serialNumber, stateModel);
+						dataModel.latestStates
+							.computeIfAbsent(
+								record.serialNumber,
+								k -> new ArrayList<>()
+							)
+							.add(stateModel);
 						stateUpdates.add(record.serialNumber);
 					} catch (JsonSyntaxException e) {
 						logger.error(
@@ -423,7 +433,7 @@ public class Modeler implements Runnable {
 			logger.debug("Removed some wifi scan entries from data model");
 		}
 		if (
-			dataModel.latestState.entrySet()
+			dataModel.latestStates.entrySet()
 				.removeIf(e -> !isRRMEnabled(e.getKey()))
 		) {
 			logger.debug("Removed some state entries from data model");
--- a/src/main/java/com/facebook/openwifirrm/modules/Modeler.java
+++ b/src/main/java/com/facebook/openwifirrm/modules/Modeler.java
@@ -9,7 +9,6 @@
 package com.facebook.openwifirrm.modules;
 
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
@@ -270,10 +269,10 @@ public class Modeler implements Runnable {
 			if (state != null) {
 				try {
 					State stateModel = gson.fromJson(state, State.class);
-					dataModel.latestStates.put(
+					dataModel.latestStates.computeIfAbsent(
 						device.serialNumber,
-						new ArrayList<>(Arrays.asList(stateModel))
-					);
+						k -> new ArrayList<>()
+					).add(stateModel);
 					logger.debug(
 						"Device {}: added initial state from uCentralGw",
 						device.serialNumber
--- a/src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
+++ b/src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
@@ -423,8 +423,7 @@ public class ModelerUtils {
 					if (association == null) {
 						continue;
 					}
-					String key = String.format(
-						"bssid: %s, station: %s",
+					String key = getBssidStationKeyPair(
 						association.bssid,
 						association.station
 					);
@@ -434,16 +433,21 @@ public class ModelerUtils {
 					AggregatedState aggState =
 						new AggregatedState(association, radioInfo);
 
-					boolean toBeAggregated = false;
+					/**
+					 * Indicate if the aggState can be merged into some old AggregatedState.
+					 * If true, it will be merged by appending its mcs/rssi field to the old one.
+					 * If false, it will be added to the list aggregatedStates.
+					*/
+					boolean canBeMergedToOldAggregatedState = false;
 					for (
 						AggregatedState oldAggregatedState : aggregatedStates
 					) {
 						if (oldAggregatedState.add(aggState)) {
-							toBeAggregated = true;
+							canBeMergedToOldAggregatedState = true;
 							break;
 						}
 					}
-					if (!toBeAggregated) {
+					if (!canBeMergedToOldAggregatedState) {
 						aggregatedStates.add(aggState);
 					}
 					bssidToAggregatedStates.put(key, aggregatedStates);
@@ -517,7 +521,6 @@ public class ModelerUtils {
 				addStateToAggregation(bssidToAggregatedStates, state);
 			}
 		}
-
 		return aggregatedStates;
 	}
 
@@ -543,5 +546,14 @@ public class ModelerUtils {
 			}
 		}
 		return latestState;
+	}
+
+	/**  Create a key pair consisted of bssid and station string */
+	public static String getBssidStationKeyPair(String bssid, String station) {
+		return String.format(
+			"bssid: %s, station: %s",
+			bssid,
+			station
+		);
 	}
 }
--- a/src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
+++ b/src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
@@ -525,7 +525,7 @@ public class ModelerUtils {
 	}
 
 	/**
-	 * This method gets the most latest State from latestStates per device.
+	 * This method gets the most recent State from latestStates per device.
 	 *
 	 * @param latestStates list of latest States per device.
 	 * @return Map<String, State> a map from device String to latest State.
--- a/src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
+++ b/src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
@@ -510,7 +510,7 @@ public class ModelerUtils {
 			for (State state : states) {
 				if (refTimeMs - state.unit.localtime > obsoletionPeriodMs) {
 					// discard obsolete entries
-					continue;
+					break;
 				}
 
 				addStateToAggregation(bssidToAggregatedStates, state);
@@ -535,7 +535,10 @@ public class ModelerUtils {
 		) {
 			String key = stateEntry.getKey();
 			List<State> value = stateEntry.getValue();
-			latestState.put(key, value.get(value.size() - 1));
+			if(value.isEmpty()){
+				latestState.put(key, null);
+			}
+			else{latestState.put(key, value.get(value.size() - 1));}
 		}
 		return latestState;
 	}
--- a/src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
+++ b/src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
@@ -387,7 +387,8 @@ public class ModelerUtils {
 		return aggregatedWifiScans;
 	}
 
-	/** This method converts the input State info to an AggregatedState
+	/**
+	 * This method converts the input State info to an AggregatedState
 	 * and adds it to the bssidToAggregatedStates map. If the bssid&station
 	 * of the input State does not exist in the map, create a new
 	 * AggregatedState list. If the bssid&station of the input State exists,
@@ -408,10 +409,14 @@ public class ModelerUtils {
 				continue;
 			}
 			for (SSID ssid : stateInterface.ssids) {
-
-				int[] radios = new int[] { ssid.radio.get("channel").getAsInt(),
-					ssid.radio.get("channel_width").getAsInt(),
-					ssid.radio.get("tx_power").getAsInt() };
+				Map<String, Integer> radioInfo = new HashMap<>();
+				radioInfo.put("channel", ssid.radio.get("channel").getAsInt());
+				radioInfo.put(
+					"channel_width",
+					ssid.radio.get("channel_width").getAsInt()
+				);
+				radioInfo
+					.put("tx_power", ssid.radio.get("tx_power").getAsInt());
 
 				for (Association association : ssid.associations) {
 					if (association == null) {
@@ -426,7 +431,7 @@ public class ModelerUtils {
 						bssidToAggregatedStates
 							.computeIfAbsent(key, k -> new ArrayList<>());
 					AggregatedState aggState =
-						new AggregatedState(association, radios);
+						new AggregatedState(association, radioInfo);
 
 					boolean toBeAggregated = false;
 					for (
@@ -515,9 +520,11 @@ public class ModelerUtils {
 		return aggregatedStates;
 	}
 
-	/** This method gets the most latest State from latestStates per device.
+	/**
+	 * This method gets the most latest State from latestStates per device.
+	 *
 	 * @param latestStates list of latest States per device.
-	 * @return Map<String, State>  a map from device String to  most latest State.
+	 * @return Map<String, State> a map from device String to latest State.
 	 */
 	public static Map<String, State> getLatestState(
 		Map<String, List<State>> latestStates
--- a/src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
+++ b/src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
@@ -13,6 +13,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.stream.Collectors;
 
 import org.slf4j.Logger;
@@ -529,16 +530,17 @@ public class ModelerUtils {
 	public static Map<String, State> getLatestState(
 		Map<String, List<State>> latestStates
 	) {
-		Map<String, State> latestState = new HashMap<>();
+		Map<String, State> latestState = new ConcurrentHashMap<>();
 		for (
 			Map.Entry<String, List<State>> stateEntry : latestStates.entrySet()
 		) {
 			String key = stateEntry.getKey();
 			List<State> value = stateEntry.getValue();
-			if(value.isEmpty()){
+			if (value.isEmpty()) {
 				latestState.put(key, null);
+			} else {
+				latestState.put(key, value.get(value.size() - 1));
 			}
-			else{latestState.put(key, value.get(value.size() - 1));}
 		}
 		return latestState;
 	}
--- a/src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
+++ b/src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
@@ -24,6 +24,11 @@ import com.facebook.openwifirrm.modules.Modeler.DataModel;
 import com.facebook.openwifirrm.ucentral.WifiScanEntry;
 import com.facebook.openwifirrm.ucentral.informationelement.HTOperation;
 import com.facebook.openwifirrm.ucentral.informationelement.VHTOperation;
+import com.facebook.openwifirrm.ucentral.models.AggregatedState;
+import com.facebook.openwifirrm.ucentral.models.State;
+import com.facebook.openwifirrm.ucentral.models.State.Interface;
+import com.facebook.openwifirrm.ucentral.models.State.Interface.SSID;
+import com.facebook.openwifirrm.ucentral.models.State.Interface.SSID.Association;
 
 /**
  * Modeler utilities.
@@ -380,5 +385,151 @@ public class ModelerUtils {
 			}
 		}
 		return aggregatedWifiScans;
+	}
+
+	/** This method converts the input State info to an AggregatedState
+	 * and adds it to the bssidToAggregatedStates map. If the bssid&station
+	 * of the input State does not exist in the map, create a new
+	 * AggregatedState list. If the bssid&station of the input State exists,
+	 * then convert State to AggregatedState and check if there exits an
+	 * AggregatedState of the same radio. If there does, append the value
+	 * of aggregation field to the existing AggregatedState, if not, create
+	 * a new AggregatedState and add it to the list.
+	 *
+	 * @param bssidToAggregatedStates map from bssid&station to a list of AggregatedState.
+	 * @param state the state that is to be added.
+	 */
+	static void addStateToAggregation(
+		Map<String, List<AggregatedState>> bssidToAggregatedStates,
+		State state
+	) {
+		for (Interface stateInterface : state.interfaces) {
+			if (stateInterface.ssids == null) {
+				continue;
+			}
+			for (SSID ssid : stateInterface.ssids) {
+
+				int[] radios = new int[] { ssid.radio.get("channel").getAsInt(),
+					ssid.radio.get("channel_width").getAsInt(),
+					ssid.radio.get("tx_power").getAsInt() };
+
+				for (Association association : ssid.associations) {
+					if (association == null) {
+						continue;
+					}
+					String key = String.format(
+						"bssid: %s, station: %s",
+						association.bssid,
+						association.station
+					);
+					List<AggregatedState> aggregatedStates =
+						bssidToAggregatedStates
+							.computeIfAbsent(key, k -> new ArrayList<>());
+					AggregatedState aggState =
+						new AggregatedState(association, radios);
+
+					boolean toBeAggregated = false;
+					for (
+						AggregatedState oldAggregatedState : aggregatedStates
+					) {
+						if (oldAggregatedState.add(aggState)) {
+							toBeAggregated = true;
+							break;
+						}
+					}
+					if (!toBeAggregated) {
+						aggregatedStates.add(aggState);
+					}
+					bssidToAggregatedStates.put(key, aggregatedStates);
+				}
+			}
+		}
+	}
+
+	/**
+	 * This method aggregates States by bssid&station key pair and radio info.
+	 * if two States of the same bssid&station match in channel, channel width and tx_power
+	 * need to be aggregated to one {@code AggregatedState}. Currently only mcs and
+	 * rssi fields are being aggregated. They are of List<Integer> type in AggregatedState,
+	 * which list all the values over the time.
+	 *
+	 * @param dataModel the data model which includes the latest recorded States.
+	 * @param obsoletionPeriodMs the maximum amount of time (in milliseconds) it
+	 *                           is worth aggregating over, starting from the
+	 *                           most recent States and working backwards in time.
+	 * 							 A State exactly {@code obsoletionPeriodMs} ms earlier
+	 * 							 than the most recent State is considered non-obsolete
+	 *                           (i.e., the "non-obsolete" window is inclusive).
+	 *                           Must be non-negative.
+	 * @param refTimeMs	the reference time were passed to make testing easier
+	 * @return  Map<String, Map<String, List<AggregatedState>>> A map from serial number to
+	 * 							a map from bssid_station String pair to a list of AggregatedState.
+	 */
+	public static Map<String, Map<String, List<AggregatedState>>> getAggregatedStates(
+		Modeler.DataModel dataModel,
+		long obsoletionPeriodMs,
+		long refTimeMs
+	) {
+		if (obsoletionPeriodMs < 0) {
+			throw new IllegalArgumentException(
+				"obsoletionPeriodMs must be non-negative."
+			);
+		}
+		Map<String, Map<String, List<AggregatedState>>> aggregatedStates =
+			new HashMap<>();
+
+		for (
+			Map.Entry<String, List<State>> deviceToStateList : dataModel.latestStates
+				.entrySet()
+		) {
+			String serialNumber = deviceToStateList.getKey();
+			List<State> states = deviceToStateList.getValue();
+
+			if (states.isEmpty()) {
+				continue;
+			}
+
+			/**
+			 * Sort in reverse chronological order. Sorting is done just in case the
+			 * States in the original list are not chronological already - although
+			 * they are inserted chronologically, perhaps latency, synchronization, etc.
+			 */
+			states.sort(
+				(state1, state2) -> -Long.compare(state1.unit.localtime, state2.unit.localtime)
+			);
+
+			Map<String, List<AggregatedState>> bssidToAggregatedStates =
+				aggregatedStates
+					.computeIfAbsent(serialNumber, k -> new HashMap<>());
+
+			for (State state : states) {
+				if (refTimeMs - state.unit.localtime > obsoletionPeriodMs) {
+					// discard obsolete entries
+					continue;
+				}
+
+				addStateToAggregation(bssidToAggregatedStates, state);
+			}
+		}
+
+		return aggregatedStates;
+	}
+
+	/** This method gets the most latest State from latestStates per device.
+	 * @param latestStates list of latest States per device.
+	 * @return Map<String, State>  a map from device String to  most latest State.
+	 */
+	public static Map<String, State> getLatestState(
+		Map<String, List<State>> latestStates
+	) {
+		Map<String, State> latestState = new HashMap<>();
+		for (
+			Map.Entry<String, List<State>> stateEntry : latestStates.entrySet()
+		) {
+			String key = stateEntry.getKey();
+			List<State> value = stateEntry.getValue();
+			latestState.put(key, value.get(value.size() - 1));
+		}
+		return latestState;
 	}
 }
--- a/src/main/java/com/facebook/openwifirrm/optimizers/channel/LeastUsedChannelOptimizer.java
+++ b/src/main/java/com/facebook/openwifirrm/optimizers/channel/LeastUsedChannelOptimizer.java
@@ -376,8 +376,7 @@ public class LeastUsedChannelOptimizer extends ChannelOptimizer {
 				}
 
 				// Get current channel of the device
-				List<State> states = model.latestStates.get(serialNumber);
-				State state = states.get(states.size() - 1);
+				State state = latestState.get(serialNumber);
 				if (state == null) {
 					logger.debug(
 						"Device {}: No state found, skipping...",
--- a/src/main/java/com/facebook/openwifirrm/optimizers/channel/LeastUsedChannelOptimizer.java
+++ b/src/main/java/com/facebook/openwifirrm/optimizers/channel/LeastUsedChannelOptimizer.java
@@ -21,6 +21,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.facebook.openwifirrm.DeviceDataManager;
+import com.facebook.openwifirrm.modules.ModelerUtils;
 import com.facebook.openwifirrm.modules.Modeler.DataModel;
 import com.facebook.openwifirrm.ucentral.UCentralConstants;
 import com.facebook.openwifirrm.ucentral.UCentralUtils;
@@ -340,8 +341,10 @@ public class LeastUsedChannelOptimizer extends ChannelOptimizer {
 				AVAILABLE_CHANNELS_BAND
 			);
 
+		Map<String, State> latestState =
+			ModelerUtils.getLatestState(model.latestStates);
 		Map<String, String> bssidsMap =
-			UCentralUtils.getBssidsMap(model.latestState);
+			UCentralUtils.getBssidsMap(latestState);
 
 		for (String band : bandsMap.keySet()) {
 			// Performance metrics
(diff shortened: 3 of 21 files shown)


src/main/java/com/facebook/openwifirrm/modules/Modeler.java
```diff
@@ -93,8 +95,9 @@ public static class DataModel {
 		public Map<String, List<List<WifiScanEntry>>> latestWifiScans =
 			new ConcurrentHashMap<>();
 
-		/** List of latest state per device. */
-		public Map<String, State> latestState = new ConcurrentHashMap<>();
+		/** List of latest states per device. */
+		public Map<String, List<State>> latestStates =
```
RockyMandayam2, line 99: I think these two lines can be one line? Not sure if it exceeds 80 chars

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
```diff
@@ -299,8 +305,12 @@ private void processData(InputData data) {
...
 						State stateModel = gson.fromJson(state, State.class);
-						dataModel.latestState
-							.put(record.serialNumber, stateModel);
+						dataModel.latestStates
+							.computeIfAbsent(
+								record.serialNumber,
+								k -> new ArrayList<>()
+							)
+							.add(stateModel);
 						stateUpdates.add(record.serialNumber);
```
RockyMandayam2, line 314: Similar to how for wifiscans, we use `params.wifiScanBufferSize`, can we create and use something like `params.stateBufferSize` and make sure not to exceed the buffer size?

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -381,4 +386,150 @@ public static Map<String, Map<String, WifiScanEntry>> getAggregatedWifiScans(
 		}
 		return aggregatedWifiScans;
 	}
+
+	/** This method converts the input State info to an AggregatedState
```
RockyMandayam2, line 390: Start javadoc on a separate line

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -381,4 +386,150 @@ public static Map<String, Map<String, WifiScanEntry>> getAggregatedWifiScans(
...
+	 */
+	static void addStateToAggregation(
+		Map<String, List<AggregatedState>> bssidToAggregatedStates,
+		State state
+	) {
+		for (Interface stateInterface : state.interfaces) {
+			if (stateInterface.ssids == null) {
+				continue;
+			}
+			for (SSID ssid : stateInterface.ssids) {
```
RockyMandayam2, line 410: Maybe more just personal preference: I think it would make sense to take this second for-loop and move it to its own method (which takes in `bssidToAggregatedStates` and `stateInterface`)? Mainly just to try and make this method less "nested"

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -381,4 +386,150 @@ public static Map<String, Map<String, WifiScanEntry>> getAggregatedWifiScans(
...
+
+				int[] radios = new int[] { ssid.radio.get("channel").getAsInt(),
```
RockyMandayam2, line 412: Seems like this var should be called something like `radioInfo` instead of `radios`

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -381,4 +386,150 @@ public static Map<String, Map<String, WifiScanEntry>> getAggregatedWifiScans(
...
+					ssid.radio.get("channel_width").getAsInt(),
+					ssid.radio.get("tx_power").getAsInt() };
```
RockyMandayam2, line 414: I prefer the `};` on a separate line

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -381,4 +386,150 @@ public static Map<String, Map<String, WifiScanEntry>> getAggregatedWifiScans(
...
+				}
+
+				addStateToAggregation(bssidToAggregatedStates, state);
+			}
+		}
+
+		return aggregatedStates;
+	}
+
+	/** This method gets the most latest State from latestStates per device.
```
RockyMandayam2, line 518: Start javadoc on the next line.

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -381,4 +386,150 @@ public static Map<String, Map<String, WifiScanEntry>> getAggregatedWifiScans(
...
+	 * @param latestStates list of latest States per device.
```
RockyMandayam2, line 519: Add new line between the text above and the params

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -381,4 +386,150 @@ public static Map<String, Map<String, WifiScanEntry>> getAggregatedWifiScans(
...
+	 * @return Map<String, State>  a map from device String to  most latest State.
```
RockyMandayam2, line 520: delete extra space before `a map` and change `most latest` to just `latest` (latest already means most recent)

src/main/java/com/facebook/openwifirrm/optimizers/channel/LeastUsedChannelOptimizer.java
```diff
@@ -373,7 +376,8 @@ public Map<String, Map<String, Integer>> computeChannelMap() {
 				}
 
 				// Get current channel of the device
-				State state = model.latestState.get(serialNumber);
+				List<State> states = model.latestStates.get(serialNumber);
+				State state = states.get(states.size() - 1);
```
RockyMandayam2, line 380: You created a `latestState` earlier in this method - you could use that and replace these two lines with: `State state = latestState.get(serialNumber)`

src/main/java/com/facebook/openwifirrm/optimizers/tpc/LocationBasedOptimalTPC.java
```diff
@@ -175,7 +175,9 @@ private void buildTxPowerMapForChannel(
 		// Filter out the invalid APs (e.g., no radio, no location data)
 		// Update txPowerChoices, boundary, apLocX, apLocY for the optimization
 		for (String serialNumber : serialNumbers) {
-			State state = model.latestState.get(serialNumber);
+
```
RockyMandayam2, line 178: Remove this new line

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
```diff
@@ -0,0 +1,214 @@
...
+package com.facebook.openwifirrm.ucentral.models;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+import com.facebook.openwifirrm.ucentral.models.State.Interface.SSID.Association;
+import com.facebook.openwifirrm.ucentral.models.State.Interface.SSID.Association.Rate;
+
+public class AggregatedState {
```
RockyMandayam2, line 18: Add a javadoc for all top-level classes

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
```diff
@@ -0,0 +1,214 @@
...
+
+	public class AggregatedRate {
```
RockyMandayam2, line 20: Make this class `static`

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
```diff
@@ -0,0 +1,214 @@
...
+			int size = rates.size();
+			if (size == 0) {
+				return;
+			}
+			add(rates);
+		}
+
+		/** Add a list of Rates to the AggregatedRate */
+		public void add(List<Rate> rates) {
+			if (rates.isEmpty()) {
```
RockyMandayam2, line 41: You do not need this check, the code will just not enter the for loop body if there are no elements.

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
```diff
@@ -0,0 +1,214 @@
...
+				return;
+			}
+			for (Rate rate : rates) {
```
RockyMandayam2, line 44: Without the above check, this code is simple enough that you could just move it to the constructor (I don't have a preference).

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
```diff
@@ -0,0 +1,214 @@
...
+
+		/** Constructor with no args */
+		public AggregatedRate() {
+			this.mcs = new ArrayList<>();
+		}
+
+		/** Constructor */
+		public AggregatedRate(List<Rate> rates) {
+			int size = rates.size();
+			if (size == 0) {
```
RockyMandayam2, line 33: You do not need this check (but if you did, better to use `isEmpty()` )

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
```diff
@@ -0,0 +1,214 @@
...
+			add(rates);
+		}
+
+		/** Add a list of Rates to the AggregatedRate */
+		public void add(List<Rate> rates) {
+			if (rates.isEmpty()) {
+				return;
+			}
+			for (Rate rate : rates) {
+				this.mcs.add(rate.mcs);
```
RockyMandayam2, line 45: Minor: You use `this.mcs` in some places and `mcs` in some places. I think the norm is not to use `this.` unless it's needed except maybe in constructors...

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
```diff
@@ -0,0 +1,214 @@
...
+
+			Radio other = (Radio) obj;
+			return channel == other.channel &&
+				channel_width == other.channel_width &&
+				tx_power == other.tx_power;
+		}
+	}
+
+	@Override
+	public int hashCode() {
```
RockyMandayam2, line 136: Add `equals()` or remove `hashCode()` (not a good idea to have one and not the other)

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
RockyMandayam2, line 18: I remember there was a desire for not just lists of stats but also aggregated values (like averages). Is that supported?

RockyMandayam2 (changes requested):
I haven't yet gotten to the weeds of the aggregation logic, but I don't want to keep you waiting too long for comments. Here are my first set of comments.

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
```diff
@@ -0,0 +1,214 @@
...
+
+		public Radio() {}
+
+		public Radio(int channel, int channel_width, int tx_power) {
+			this.channel = channel;
+			this.channel_width = channel_width;
+			this.tx_power = tx_power;
+		}
+
+		public Radio(int[] radios) {
```
RockyMandayam2, line 105: Seems like something like `radioInfo` would be a better name than `radios`?

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
```diff
@@ -0,0 +1,214 @@
...
+
+	/** Construct from Aggregatedstate */
+	public AggregatedState(AggregatedState state) {
+		this.rx_rate = new AggregatedRate();
+		this.tx_rate = new AggregatedRate();
+		this.rssi = new ArrayList<>();
+		add(state);
+	}
+
+	/** Construt from Asscociation and radio */
```
RockyMandayam2, line 156: `Construt` -> `Construct`

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
```diff
@@ -0,0 +1,214 @@
...
+	public AggregatedState(Association association, int[] radios) {
```
RockyMandayam2, line 157: Same - I think something like `radioInfos` is a better name than `radios`?

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
```diff
@@ -0,0 +1,214 @@
...
+		this.tx_failed = association.tx_failed;
+		this.tx_packets = association.tx_packets;
+		this.tx_rate.add(association.tx_rate);
+		this.tx_retries = association.tx_retries;
+		this.ack_signal = association.ack_signal;
+		this.ack_signal_avg = association.ack_signal_avg;
+		this.radio = new Radio(radios);
+	}
+
+	/** Add an AggregatedState to this AggregatedState. Succeed only when the two
```
RockyMandayam2, line 181: Start javadoc on next line

src/test/java/com/facebook/openwifirrm/modules/ModelerUtilsTest.java
```diff
@@ -547,4 +552,278 @@ void testPostDot11nAggregatedWifiScanEntry() {
...
+			2,
+			40,
+			2000,
+			bssidA,
+			new String[] { "stationA2", "stationA2" },
+			new int[] { 180, 67 },
+			TestUtils.DEFAULT_LOCAL_TIME - 800
+		);
+
+		//As State time3ToStateA is obsolescent, it should not be aggregated.
```
RockyMandayam2, line 719: `obsolescent` -> `obsolete`

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
```diff
@@ -0,0 +1,214 @@
...
+	public long tx_duration;
+	public long tx_failed;
+	public long tx_packets;
+	public AggregatedRate tx_rate;
+	public long tx_retries;
+	public int ack_signal;
+	public int ack_signal_avg;
+	public Radio radio;
+
+	public class Radio {
```
RockyMandayam2, line 92: Make this class `static`

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
```diff
@@ -267,7 +270,10 @@ private void fetchInitialData() {
 			if (state != null) {
 				try {
 					State stateModel = gson.fromJson(state, State.class);
-					dataModel.latestState.put(device.serialNumber, stateModel);
+					dataModel.latestStates.put(
+						device.serialNumber,
+						new ArrayList<>(Arrays.asList(stateModel))
```
wuweicaifb, line 275: 1. I don't think `new ArrayList<>` is needed here
2. Don't you want to append the `latestStates` list if `serialNumber` already exists? like line 308-313.

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
wuweicaifb, line 412: Instead of an int array, a new struct with `channel`, `channel_width` and `tx_power` would be more self-explanatory.

wuweicaifb (commented):
Two general comment:
1. I found the some code style did not align with other code. Could you run `mvn spotless:apply` to reformat the code?
2. Maybe it's hard to split the code changes into smaller PRs for this change, but smaller PR would be much easier for review and iterate.

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
jwoos, line 99: this needs to be thread safe. we need to use thread safe structures and/or proper locking. As is there will be race conditions and weird behaviors (i don't know the full behavior of java but i imagine it could crash given certain code paths).

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
```diff
@@ -0,0 +1,214 @@
...
+		this.rx_packets = association.rx_packets;
+		this.rx_rate.add(association.rx_rate);
+		this.tx_bytes = association.tx_bytes;
+		this.tx_duration = association.tx_duration;
+		this.tx_failed = association.tx_failed;
+		this.tx_packets = association.tx_packets;
+		this.tx_rate.add(association.tx_rate);
+		this.tx_retries = association.tx_retries;
+		this.ack_signal = association.ack_signal;
+		this.ack_signal_avg = association.ack_signal_avg;
```
jwoos, line 177: camel case? the original structure is only snake case so that the json gets parsed appropriately

jwoos (changes requested):
Since Rocky and Wuwei left a lot of the comments I would have, I'll focus more on the main problem I see - which is with thread safety.

The original implementation used a `ConcurrentHashMap` and since it only dealt with a single item it was fine. But the new implementation deals with a nested data structure so I imagine there can be some issues (I'm not sure about the behavior of ConcurrentHashMap but I imagine it doesn't guarantee safe access to the underlying elements since no thread safe hash map I've seen does - i.e. if you were to access the list concurrently it may be an issue). We need to lock the internal structure or make it use a thread safe structure as well (which may or may not be enough by itself)

src/test/java/com/facebook/openwifirrm/modules/ModelerUtilsTest.java
```diff
@@ -547,4 +552,278 @@ void testPostDot11nAggregatedWifiScanEntry() {
...
+		assertEquals(
+			aggregatedMap2.get(serialNumberB).get(TestUtils.getBssidStationKeyPair(bssidB, "stationB")).get(0).rssi, Arrays.asList(-30)
+		);
+
+		assertEquals(
+			aggregatedMap2.get(serialNumberC).get(TestUtils.getBssidStationKeyPair(bssidC, "stationC")).get(0).rssi, Arrays.asList(-100)
+		);
+
+		assertEquals(
+			aggregatedMap2.get(serialNumberA).get(TestUtils.getBssidStationKeyPair(bssidA, "stationA1")).size(), aggregatedMap.get(serialNumberA).get(TestUtils.getBssidStationKeyPair(bssidA, "stationA1")).size()
```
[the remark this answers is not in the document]
zhiqiand: Yes. Actually no thing changed after running `mvn spotless:apply` again

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
RockyMandayam2: I mean a new state buffer size is needed. We don't want to store all states and never delete them!

src/test/java/com/facebook/openwifirrm/modules/ModelerUtilsTest.java
```diff
@@ -547,4 +551,290 @@ void testPostDot11nAggregatedWifiScanEntry() {
...
+			bssidC,
+			stationC,
+			new int[] { 100, 200, 300 }
+		);
+
+		Map<String, List<AggregatedState>> bssidToAggregatedStates =
+			new HashMap<>();
+		bssidToAggregatedStates.put(
+			TestUtils.getBssidStationKeyPair(bssidA, stationA1),
+			new ArrayList<>(Arrays.asList(aggStateA1))
```
wuweicaifb, line 602: remove `new ArrayList<>`. Same for the following

src/test/java/com/facebook/openwifirrm/modules/ModelerUtilsTest.java
```diff
@@ -547,4 +551,290 @@ void testPostDot11nAggregatedWifiScanEntry() {
...
+		);
+
+		Map<String, Map<String, List<AggregatedState>>> aggregatedMap =
+			ModelerUtils
+				.getAggregatedStates(dataModel, obsoletionPeriodMs, refTimeMs);
+
+		assertEquals(
+			aggregatedMap.get(serialNumberA).get(TestUtils.getBssidStationKeyPair(bssidA, "stationA1")).size(),
+			2
+		);
```
wuweicaifb, line 754: Did you run the formatter? This line is obviously over 80 chars.

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
```diff
@@ -0,0 +1,234 @@
...
+		public AggregatedRate() {
+			mcs = new ArrayList<>();
+		}
+
+		/** Add a Rate to the AggregatedRate */
+		public void add(Rate rate) {
+			if (rate == null) {
+				return;
+			}
+			if (mcs == null || mcs.isEmpty()) {
```
wuweicaifb, line 39: Can `mcs` ever become null? If so, you need to initialize it before line 43 `mcs.add(rate.mcs)`

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
```diff
@@ -0,0 +1,234 @@
...
+		public Radio() {}
+
+		public Radio(int channel, int channelWidth, int txPower) {
+			this.channel = channel;
+			this.channelWidth = channelWidth;
+			this.txPower = txPower;
+		}
+
+		public Radio(Map<String, Integer> radioInfo) {
+			channel = radioInfo.getOrDefault("channel", null);
```
wuweicaifb, line 91: You cannot assign null to primitive types like int. Please either box it or use -1 (or other feasible "invalid" number) as default. Same for `channelWidth` and `txPower`

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
```diff
@@ -0,0 +1,234 @@
...
+			txFailed == other.txFailed && txPackets == other.txPackets &&
+			txRate.bitRate == other.txRate.bitRate &&
+			txRate.chWidth == other.txRate.chWidth && txRate.mcs
+				.equals(other.txRate.mcs) &&
+			txRetries == other.txRetries && ackSignal == other.ackSignal &&
+			ackSignalAvg == other.ackSignalAvg && radio.equals(other.radio);
+	}
+
+	/** Constructor with no args */
+	public AggregatedState() {
```
wuweicaifb, line 157: Recommend putting constructors before other public methods.

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
```diff
@@ -0,0 +1,234 @@
...
+import java.util.Map;
+import java.util.Objects;
+
+import com.facebook.openwifirrm.ucentral.models.State.Interface.SSID.Association;
+import com.facebook.openwifirrm.ucentral.models.State.Interface.SSID.Association.Rate;
+
+/**
+ * Aggregation model for State aggregation. Only contains info useful for analysis.
+ */
+public class AggregatedState {
```
wuweicaifb, line 22: Please rearrange classes, members and methods in this classes to align with the styles with others.

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
```diff
@@ -0,0 +1,234 @@
...
+		this.rssi = new ArrayList<>();
+		this.radio = new Radio();
+	}
+
+	/** Construct from Aggregatedstate */
+	public AggregatedState(AggregatedState state) {
+		this.rxRate = new AggregatedRate();
+		this.txRate = new AggregatedRate();
+		this.rssi = new ArrayList<>();
+		add(state);
```
wuweicaifb, line 169: Doesn't it always return false because `state.hashCode()` is different from `hashCode(null,null,null)`?

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -381,4 +386,157 @@ public static Map<String, Map<String, WifiScanEntry>> getAggregatedWifiScans(
...
+	 */
+	public static Map<String, State> getLatestState(
+		Map<String, List<State>> latestStates
+	) {
+		Map<String, State> latestState = new HashMap<>();
+		for (
+			Map.Entry<String, List<State>> stateEntry : latestStates.entrySet()
+		) {
+			String key = stateEntry.getKey();
+			List<State> value = stateEntry.getValue();
```
wuweicaifb, line 537: I wonder if it's possible that the list is empty here.

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -381,4 +386,157 @@ public static Map<String, Map<String, WifiScanEntry>> getAggregatedWifiScans(
...
+
+			/**
+			 * Sort in reverse chronological order. Sorting is done just in case the
+			 * States in the original list are not chronological already - although
+			 * they are inserted chronologically, perhaps latency, synchronization, etc.
+			 */
+			states.sort(
+				(state1, state2) -> -Long.compare(state1.unit.localtime, state2.unit.localtime)
+			);
+
```
wuweicaifb, line 505: So the first one rather than the last one should be latest, right? but you retrieve the last one as latest in all other places.

wuweicaifb (changes requested):

src/test/java/com/facebook/openwifirrm/modules/ModelerUtilsTest.java
zhiqiand: Yes I have run the formatter but it doesn't change anything. Do I need to format it manually?

wuweicaifb: Seems that `spotless` does not enforce the line width? @elludraon  If so, I think there's no need to do it manually. but I wonder if your IDE or editor could do it for you.

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
RockyMandayam2: The first one is the most recent. And she starts iterating from the beginning, so it's fine, right?

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -381,4 +386,157 @@ public static Map<String, Map<String, WifiScanEntry>> getAggregatedWifiScans(
...
+			Map<String, List<AggregatedState>> bssidToAggregatedStates =
+				aggregatedStates
+					.computeIfAbsent(serialNumber, k -> new HashMap<>());
+
+			for (State state : states) {
+				if (refTimeMs - state.unit.localtime > obsoletionPeriodMs) {
+					// discard obsolete entries
+					continue;
```
RockyMandayam2, line 513: Can you `break` instead since all the next entries will be obsolete?

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
wuweicaifb: But `getLatestState` in this file selected the last one as "latest".

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
zhiqiand: Maybe I need to add one more condition check in add(state)

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
```diff
@@ -0,0 +1,236 @@
...
+
+		/** Add a Rate to the AggregatedRate */
+		public void add(Rate rate) {
+			if (rate == null) {
+				return;
+			}
+			if (mcs == null || mcs.isEmpty()) {
+				bitRate = rate.bitrate;
+				chWidth = rate.chwidth;
+				mcs = new ArrayList<>();
```
wuweicaifb, line 43: nit: only `new ArrayList<>()` when it's null.

wuweicaifb (approved):
Approved to unblocked you from my side. Please wait for other reviwers' reponse too.

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
```diff
@@ -305,13 +305,18 @@ private void processData(InputData data) {
...
 							.computeIfAbsent(
 								record.serialNumber,
 								k -> Collections
 									.synchronizedList(new ArrayList<>())
-							)
-							.add(stateModel);
+							);
+						if (latestStatesList.size() >= params.stateBufferSize) {
+							latestStatesList.remove(0);
+						}
```
jwoos, line 316: is it okay to just remove?

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
```diff
@@ -271,7 +272,7 @@ private void fetchInitialData() {
 					State stateModel = gson.fromJson(state, State.class);
 					dataModel.latestStates.computeIfAbsent(
 						device.serialNumber,
-						k -> new ArrayList<>()
+						k -> Collections.synchronizedList(new ArrayList<>())
```
jwoos, line 275: AFAICT, probably the best option. COW is going to be too expensive.

jwoos: also maybe not in this PR but i think there are no accesses besides first and last usually (besides sorting the whole thing) - linked list might make more sense vs allocating contiguous block

jwoos (approved):
looks good, pending one small comment

src/test/java/com/facebook/openwifirrm/optimizers/TestUtils.java
```diff
@@ -644,7 +696,77 @@ public static State createState(
...
+			new int[] { channelWidthA, channelWidthB },
+			new int[] { txPowerA, txPowerB },
+			new String[] { bssidA, bssidB },
+			new String[][] { stationsA, stationsB },
+			new int[][] { clientRssisA, clientRssisB },
+			localtime
+		);
+	}
+
+	public static String getBssidStationKeyPair(String bssid, String station) {
```
RockyMandayam2, line 765: Since you do have a method for this, maybe move it somewhere else (`ModelerUtils`? Or `UCentralUtils`?) and use it in `ModelerUtils` instead of `String.format` there?

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
```diff
@@ -0,0 +1,241 @@
...
+	 * Add an AggregatedState to this AggregatedState. Succeed only when the two
+	 * matches in hashCode.
+	 *
+	 * @param state input AggregatedState
+	 * @return boolean true if the two matches in bssid, station, channel,
+	 *         channel_width and tx_power
+	 */
+	public boolean add(AggregatedState state) {
+		if (
+			(bssid == null && station == null && radio == null) ||
```
RockyMandayam2, line 213: I don't see a `null` ever passed in, except potentially through the constructor from `addStateToAggregation`. If that could be null there, I think we should do a null-check there (like `if (association.bssid == null) {continue;}`) and then remove the null check here. Same with `station` and `radio`. Then we can have a guarantee that `bssid`, `station`, and `radio` are never null (and that would be good to put in their javadoc)

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
```diff
@@ -0,0 +1,241 @@
...
+	/** Constructor with no args */
+	public AggregatedState() {
+		this.rxRate = new AggregatedRate();
+		this.txRate = new AggregatedRate();
+		this.rssi = new ArrayList<>();
+		this.radio = new Radio();
+	}
+
+	/** Construct from Aggregatedstate */
+	public AggregatedState(AggregatedState state) {
```
RockyMandayam2, line 133: Is this constructor used? If not, delete.

If you keep it:
- maybe you can do just `this()` followed by `add(state)`? I think it would be "simpler" (less duplicate code, fewer lines), but the downside is the radio field would get written then overwritten
- change the javadoc's `Aggregatedstate` -> `AggregatedState` (capitalized `S`)

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -381,4 +387,161 @@ public static Map<String, Map<String, WifiScanEntry>> getAggregatedWifiScans(
...
+						association.bssid,
+						association.station
+					);
+					List<AggregatedState> aggregatedStates =
+						bssidToAggregatedStates
+							.computeIfAbsent(key, k -> new ArrayList<>());
+					AggregatedState aggState =
+						new AggregatedState(association, radioInfo);
+
+					boolean toBeAggregated = false;
```
RockyMandayam2, line 437: Maybe this is just a personal preference/misunderstanding - to me, it makes more sense to call this something like `canBeMerged` or `addedToExistingAggregateState` or something like that?

RockyMandayam2 (changes requested):
Requested some more changes. Also, please don't merge till we have the meeting with the WSEs to discuss the implementation and also to discuss if/how we might add time stamps. Thanks!

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
```diff
@@ -0,0 +1,241 @@
...
+			}
+			if (mcs == null || mcs.isEmpty()) {
+				bitRate = rate.bitRate;
+				chWidth = rate.chWidth;
+			}
+			mcs.addAll(rate.mcs);
+		}
+	}
+
+	public static class Radio {
```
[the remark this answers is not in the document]
zhiqiand: I am not sure, cuz the radio here doesn't have any aggregated fields. it serves as a key.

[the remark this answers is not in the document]
RockyMandayam2: Ok, I think just `Radio` is good

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
zhiqiand: Construct from AggregatedState, in this case `(bssid == null && station == null && radio == null)` could be true.

src/test/java/com/facebook/openwifirrm/optimizers/TestUtils.java
```diff
@@ -644,7 +694,77 @@ public static State createState(
 			new int[] { channelWidthA, channelWidthB },
 			new int[] { txPowerA, txPowerB },
 			new String[] { bssidA, bssidB },
-			new int[][] { clientRssisA, clientRssisB }
+			new String[][] {},
```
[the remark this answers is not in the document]
zhiqiand: `clientRssisA` and `clientRssisB` are input parameters, it's not initialization. `new String[][] { new String[clientRssis.length] }` initialize an empty 2D string Array, if I didn't declare the size, it will run into NPE

src/test/java/com/facebook/openwifirrm/optimizers/TestUtils.java
```diff
@@ -463,15 +472,19 @@ private static State.Unit createStateUnit() {
...
 		int[] channelWidths,
 		int[] txPowers,
 		String[] bssids,
-		int[][] clientRssis
+		String[][] stations,
+		int[][] clientRssis,
+		long localtime
 	) {
 		if (
 			!(channels.length == channelWidths.length &&
```
RockyMandayam2, line 490: Check the lengths of `stations` and `clientRssis` too?

src/test/java/com/facebook/openwifirrm/optimizers/TestUtils.java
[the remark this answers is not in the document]
zhiqiand: > Ok. Above (line 624) you define the inner array size. Any particular reason for the difference?

This method is not called externally. The usage is only called by another `createState` method:
```
	public static State createState(
		int channelA,
		int channelWidthA,
		int txPowerA,
		String bssidA,
		int channelB,
		int channelWidthB,
		int txPowerB,
		String bssidB
	) {
		return createState(
			channelA,
			channelWidthA,
			txPowerA,
			bssidA,
			new int[] {},
			channelB,
			channelWidthB,
			txPowerB,
			bssidB,
			new int[] {}
		);
	}
```
You could see that actually the `clientRssisA` and `clientRssisB` are empty, so it would never try to access the first element of the empty array. I will add the size initialization as line 624 does.

src/test/java/com/facebook/openwifirrm/optimizers/TestUtils.java
```diff
@@ -463,20 +472,26 @@ private static State.Unit createStateUnit() {
...
 				channelWidths.length == txPowers.length &&
-				txPowers.length == bssids.length)
+				txPowers.length == bssids.length &&
+				channels.length == stations.length &&
```
RockyMandayam2, line 493: just personal preference, I feel like to keep consistent with the rest of the structure, `bssids.length == stations.length` is better, but up to you

src/main/java/com/facebook/openwifirrm/modules/ModelerUtils.java
```diff
@@ -381,4 +387,173 @@ public static Map<String, Map<String, WifiScanEntry>> getAggregatedWifiScans(
...
+				}
+
+				addStateToAggregation(bssidToAggregatedStates, state);
+			}
+		}
+		return aggregatedStates;
+	}
+
+	/**
+	 * This method gets the most latest State from latestStates per device.
```
RockyMandayam2, line 528: Minor: "most latest" -> "most recent"

src/test/java/com/facebook/openwifirrm/modules/ModelerUtilsTest.java
```diff
@@ -547,4 +551,289 @@ void testPostDot11nAggregatedWifiScanEntry() {
...
+
+		DataModel dataModel = new DataModel();
+
+		// This serie of StateA is used to test a valid input states.
+		State time1ToStateA = TestUtils.createState(
+			1,
+			80,
+			1000,
+			bssidA,
+			new String[] { "stationA1", "stationA2", "stationA2", "stationA3" },
```
RockyMandayam2, line 690: Can we use `final String` vars for all stations? Here and the ones later such as A4 and B. Like how it is done in `testAddStateToAggregation`

src/test/java/com/facebook/openwifirrm/modules/ModelerUtilsTest.java
```diff
@@ -547,4 +551,289 @@ void testPostDot11nAggregatedWifiScanEntry() {
...
+		final String bssidB = "bb:bb:bb:bb:bb:bb";
+		final String serialNumberC = "cccccccccccc";
+		final String bssidC = "cc:cc:cc:cc:cc:cc";
+
+		long refTimeMs = TestUtils.DEFAULT_LOCAL_TIME;
+
+		DataModel dataModel = new DataModel();
+
+		// This serie of StateA is used to test a valid input states.
+		State time1ToStateA = TestUtils.createState(
```
RockyMandayam2, line 685: I think just `time1StateA` is better (i.e., remove the `To`)? What do you think? For this and all the other similar vars.

RockyMandayam2 (changes requested):

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
```diff
@@ -0,0 +1,239 @@
...
+
+	/**
+	 * Add an AggregatedState to this AggregatedState. Succeed only when the two
+	 * matches in hashCode.
+	 *
+	 * @param state input AggregatedState
+	 * @return boolean true if the two matches in bssid, station, channel,
+	 *         channel_width and tx_power
+	 */
+	public boolean add(AggregatedState state) {
```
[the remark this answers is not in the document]
zhiqiand: You are right.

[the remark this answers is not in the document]
RockyMandayam2: Since you're planning to do a follow-up PR to store time stamps, you could resolve this in that follow-up, right? Since time stamps would make this problem irrelevant?

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
```diff
@@ -0,0 +1,239 @@
...
+			}
+			mcs.addAll(rate.mcs);
+		}
+	}
+
+	/**
+	 * Radio information with channel, channel_width and tx_power.
+	 */
+	public static class Radio {
+		public int channel;
```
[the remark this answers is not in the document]
zhiqiand: Cannot make it final, because there is a constructor with no args.

src/test/java/com/facebook/openwifirrm/optimizers/TestUtils.java
```diff
@@ -599,52 +653,84 @@ public static State createState(
...
+		int txPower,
+		String bssid,
+		String station,
+		int[] clientRssi
+	) {
+		AggregatedState state = new AggregatedState();
+		state.radio = new AggregatedState.Radio(channel, channelWidth, txPower);
+		state.bssid = bssid;
+		state.station = station;
+		for (int rssi : clientRssi) {
```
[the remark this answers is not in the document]
zhiqiand: Cannot addAll as `state.rssi` and `clientRssi ` are of different types.
```
The method addAll(Collection<? extends Integer>) in the type List<Integer> is not applicable for the arguments (int[])
```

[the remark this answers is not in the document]
RockyMandayam2: Did you try `state.rssi.addAll(clientRssi)` or `Collections.addAll(state.rssi, clientRssi)`? If the former, try the latter.

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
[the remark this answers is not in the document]
RockyMandayam2: Ok, looks like you just removed all the assignments that overrode the old values. That works.

src/test/java/com/facebook/openwifirrm/optimizers/TestUtils.java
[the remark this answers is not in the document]
zhiqiand: Yes, I tried both
```
The method addAll(Collection<? super T>, T...) in the type Collections is not applicable for the arguments (List<Integer>, int[])
```

RockyMandayam2 (approved):
Thanks!

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
```diff
@@ -299,8 +304,18 @@ private void processData(InputData data) {
...
 						State stateModel = gson.fromJson(state, State.class);
-						dataModel.latestState
-							.put(record.serialNumber, stateModel);
+						List<State> latestStatesList = dataModel.latestStates
+							.computeIfAbsent(
+								record.serialNumber,
+								k -> Collections
+									.synchronizedList(new LinkedList<>())
+							);
+						if (latestStatesList.size() >= params.stateBufferSize) {
```
elludraon, line 313: this should be a `while` loop

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
```diff
@@ -299,8 +304,18 @@ private void processData(InputData data) {
 				if (state != null) {
 					try {
 						State stateModel = gson.fromJson(state, State.class);
-						dataModel.latestState
-							.put(record.serialNumber, stateModel);
+						List<State> latestStatesList = dataModel.latestStates
+							.computeIfAbsent(
+								record.serialNumber,
+								k -> Collections
+									.synchronizedList(new LinkedList<>())
```
elludraon, line 311: Why is this `synchronizedList` ??

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
```diff
@@ -0,0 +1,221 @@
...
+		this.txRetries = association.tx_retries;
+		this.ackSignal = association.ack_signal;
+		this.ackSignalAvg = association.ack_signal_avg;
+		this.radio = new Radio(radioInfo);
+	}
+
+	@Override
+	public int hashCode() {
+		return Objects.hash(bssid, station, radio);
+	}
```
elludraon, line 175: You should NOT override `hashCode` differently from `equals`.... it will produce totally unexpected behavior unless you read this method (and if you did want to keep it, there should be a very large comment here explaining).

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
```diff
@@ -0,0 +1,221 @@
...
+	/**
+	 * Add an AggregatedState to this AggregatedState. Succeed only when the two
+	 * matches in hashCode.
+	 *
+	 * @param state input AggregatedState
+	 * @return boolean true if the two matches in bssid, station, channel,
+	 *         channel_width and tx_power
+	 */
+	public boolean add(AggregatedState state) {
+		if (hashCode() == state.hashCode()) {
```
elludraon, line 213: see above... do not override `hashCode`, just do the checks here

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
zhiqiand: I guess `if` statement also works here? Because it is in a for loop 
```
for (KafkaRecord record : data.records)
```
at each iteration we only added one `stateModel` so only one `stateModel` is to be removed.

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
zhiqiand: My intention is to make the internal structure thread-safe as well. Though a `ConcurrentMap` is already thread-safe

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
elludraon: It is possible that there are more elements in this list than should be there, the `while` loop will always truncate it to the correct size

src/main/java/com/facebook/openwifirrm/modules/Modeler.java
elludraon: Is it actually needed? It will hurt performance (always locks), whereas `ConcurrentMap` is less likely to introduce overhead.

zhiqiand: If we want it to be thread-safe. It is needed. Because the List here as a value is not guaranteed to be thread-safe.

elludraon: This is the only place actually modifying the list, right? In that case you don't need thread safety. In RRM algorithms, the whole `DataModel` is copied.

PR #88, opened 2022-09-27
Refresh token when expired

zhiqiand:
Summary:
See Task: [T127668163](https://www.internalfb.com/intern/tasks/?t=127668163)
log: 
Screen Shot 2022-09-27 at 12 23 32 PM

set `TOKEN_REFRESH_WINDOW_S` to very slow: will never hit the log to refresh access token. But print out this log in every `TOKEN_REFRESH_WINDOW_S`
Screen Shot 2022-09-27 at 12 22 27 PM

### DIFF
--- a/src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
+++ b/src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
@@ -143,6 +143,17 @@ public class UCentralClient {
 	private WebTokenResult accessToken;
 
 	/**
+	 * The unix timestamp (in seconds) keeps track of when the accessToken is created.
+	 */
+	private long created;
+
+	/**
+	 * The unix timestamp (in seconds) keeps track of last time when the accessToken 
+	 * is accessed.
+	 */
+	private long lastAccess;
+
+	/**
 	 * Constructor.
 	 * @param rrmEndpoint advertise this RRM endpoint to the SDK
 	 * @param usePublicEndpoints whether to use public or private endpoints
@@ -216,6 +227,8 @@ public class UCentralClient {
 			return false;
 		}
 		this.accessToken = token;
+		this.created = accessToken.created;
+		this.lastAccess = accessToken.created;
 		logger.info("Login successful as user: {}", username);
 		logger.debug("Access token: {}", accessToken.access_token);
 		logger.debug("Refresh token: {}", accessToken.refresh_token);
@@ -243,7 +256,7 @@ public class UCentralClient {
 		if (accessToken == null) {
 			return true;
 		}
-		return accessToken.created + accessToken.expires_in <
+		return created + accessToken.expires_in <
 			Instant.now().getEpochSecond();
 	}
 
@@ -256,7 +269,7 @@ public class UCentralClient {
 		if (accessToken == null) {
 			return true;
 		}
-		return accessToken.created + accessToken.idle_timeout <
+		return lastAccess + accessToken.idle_timeout <
 			Instant.now().getEpochSecond();
 	}
 
@@ -280,6 +293,8 @@ public class UCentralClient {
 				if (isAccessTokenTimedOut()) {
 					logger.debug("Access token timed out, refreshing the token");
 					accessToken = refreshToken();
+					created = Instant.now().getEpochSecond();
+					lastAccess = created;
 					if (accessToken != null) {
 						logger.debug("Successfully refresh token.");
 					}else{
@@ -448,6 +463,7 @@ public class UCentralClient {
 					"Authorization",
 					"Bearer " + accessToken.access_token
 				);
+				lastAccess = Instant.now().getEpochSecond();
 			}
 		} else {
 			req
@@ -501,6 +517,7 @@ public class UCentralClient {
 					"Authorization",
 					"Bearer " + accessToken.access_token
 				);
+				lastAccess = Instant.now().getEpochSecond();
 			}
 		} else {
 			req
--- a/src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
+++ b/src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
@@ -8,6 +8,7 @@
 
 package com.facebook.openwifirrm.ucentral;
 
+import java.time.Instant;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
@@ -30,6 +31,8 @@ import com.facebook.openwifirrm.ucentral.gw.models.ServiceEvent;
 import com.facebook.openwifirrm.ucentral.gw.models.StatisticsRecords;
 import com.facebook.openwifirrm.ucentral.gw.models.SystemInfoResults;
 import com.facebook.openwifirrm.ucentral.gw.models.TokenValidationResult;
+import com.facebook.openwifirrm.ucentral.gw.models.WebTokenRefreshRequest;
+import com.facebook.openwifirrm.ucentral.gw.models.WebTokenResult;
 import com.facebook.openwifirrm.ucentral.gw.models.WifiScanRequest;
 import com.facebook.openwifirrm.ucentral.prov.models.EntityList;
 import com.facebook.openwifirrm.ucentral.prov.models.InventoryTagList;
@@ -137,7 +140,10 @@ public class UCentralClient {
 	 * The access token obtained from uCentralSec, needed only when using public
 	 * endpoints.
 	 */
-	private String accessToken;
+	private WebTokenResult accessToken;
+
+	/** Time window to refresh token in seconds. */
+	private int TOKEN_REFRESH_WINDOW_S = 21600;
 
 	/**
 	 * Constructor.
@@ -184,7 +190,8 @@ public class UCentralClient {
 		Map<String, Object> body = new HashMap<>();
 		body.put("userId", username);
 		body.put("password", password);
-		HttpResponse<String> response = httpPost("oauth2", OWSEC_SERVICE, body);
+		HttpResponse<String> response =
+			httpPost("oauth2", OWSEC_SERVICE, body, null);
 		if (!response.isSuccess()) {
 			logger.error(
 				"Login failed: Response code {}, body: {}",
@@ -195,25 +202,115 @@ public class UCentralClient {
 		}
 
 		// Parse access token from response
-		JSONObject respBody;
+		WebTokenResult token;
 		try {
-			respBody = new JSONObject(response.getBody());
-		} catch (JSONException e) {
+			token = gson.fromJson(response.getBody(), WebTokenResult.class);
+		} catch (JsonSyntaxException e) {
 			logger.error("Login failed: Unexpected response", e);
 			logger.debug("Response body: {}", response.getBody());
 			return false;
 		}
-		if (!respBody.has("access_token")) {
+		if (token.access_token == null || token.access_token.isEmpty()) {
 			logger.error("Login failed: Missing access token");
-			logger.debug("Response body: {}", respBody.toString());
+			logger.debug("Response body: {}", token.toString());
 			return false;
 		}
-		this.accessToken = respBody.getString("access_token");
+		this.accessToken = token;
 		logger.info("Login successful as user: {}", username);
-		logger.debug("Access token: {}", accessToken);
+		logger.debug("Access token: {}", accessToken.access_token);
+		logger.debug("Refresh token: {}", accessToken.refresh_token);
 
 		// Load system endpoints
 		return loadSystemEndpoints();
+	}
+
+	/**
+	 * Check if an access token is expired.
+	 *
+	 * @return true if the access token is expired
+	 */
+	private boolean isAccessTokenExpired() {
+		return accessToken.created + accessToken.idle_timeout <
+			Instant.now().getEpochSecond();
+	}
+
+	/**
+	 * Check if the refresh token is expired.
+	 *
+	 * @return true if the refresh token is expired
+	 */
+	private boolean isRefreshTokenExpired() {
+		return accessToken.created + TOKEN_REFRESH_WINDOW_S <
+			Instant.now().getEpochSecond();
+	}
+
+	/**
+	 * Get access token. If the refresh token is expired, login again. 
+	 * If the access token is expired, POST a WebTokenRefreshRequest to refresh token. 
+	 * Otherwise return the current access token.
+	 *
+	 * @return a valid access token ({@code WebTokenResult})
+	 */
+	private WebTokenResult getAccessToken() {
+		if (isRefreshTokenExpired()) {
+			logger.info("Refresh token is expired, login again");
+			if (login()) {
+				return accessToken;
+			}
+			return null;
+		} else if (isAccessTokenExpired()) {
+			logger.debug("Access token is expired, start refreshing a token.");
+			WebTokenResult refresh_token = refreshToken();
+			if (refresh_token != null) {
+				logger.debug("Successfully refresh token.");
+				return refresh_token;
+			}
+			logger.error(
+				"Fail to refresh token with access token: {}",
+				accessToken.access_token
+			);
+			return null;
+		}
+		return accessToken;
+	}
+
+	/**
+	 * POST a WebTokenRefreshRequest to refresh the access token.
+	 *
+	 * @return valid access token if success, otherwise return null.
+	 */
+	private WebTokenResult refreshToken() {
+		WebTokenRefreshRequest refreshRequest = new WebTokenRefreshRequest();
+		refreshRequest.userId = username;
+		refreshRequest.refreshToken = accessToken.refresh_token;
+		logger.debug("refresh token: {}", accessToken.refresh_token);
+		Map<String, Object> query = new HashMap<>();
+		query.put("grant_type", "refresh_token");
+		HttpResponse<String> response =
+			httpPost(
+				"oauth2",
+				OWSEC_SERVICE,
+				refreshRequest,
+				query
+			);
+		if (!response.isSuccess()) {
+			logger.error(
+				"Failed to refresh token: Response code {}, body: {}",
+				response.getStatus(),
+				response.getBody()
+			);
+			return null;
+		}
+		try {
+			return gson.fromJson(response.getBody(), WebTokenResult.class);
+		} catch (JsonSyntaxException e) {
+			logger.error(
+				"Failed to serialize responseBody: Unexpected response:",
+				e
+			);
+			logger.debug("Response body: {}", response.getBody());
+			return null;
+		}
 	}
 
 	/** Read system endpoint URLs from uCentralSec. */
@@ -270,6 +367,7 @@ public class UCentralClient {
 		) {
 			return false;
 		}
+		accessToken = getAccessToken();
 		if (usePublicEndpoints && accessToken == null) {
 			return false;
 		}
@@ -284,6 +382,7 @@ public class UCentralClient {
 		if (!serviceEndpoints.containsKey(OWPROV_SERVICE)) {
 			return false;
 		}
+		accessToken = getAccessToken();
 		if (usePublicEndpoints && accessToken == null) {
 			return false;
 		}
@@ -325,7 +424,10 @@ public class UCentralClient {
 			.socketTimeout(socketTimeoutMs);
 		if (usePublicEndpoints) {
 			if (accessToken != null) {
-				req.header("Authorization", "Bearer " + accessToken);
+				req.header(
+					"Authorization",
+					"Bearer " + accessToken.access_token
+				);
 			}
 		} else {
 			req
@@ -339,37 +441,52 @@ public class UCentralClient {
 		}
 	}
 
-	/** Send a POST request with a JSON body. */
+	/** Send a POST request with a JSON body and query params. */
 	private HttpResponse<String> httpPost(
 		String endpoint,
 		String service,
-		Object body
+		Object body,
+		Map<String, Object> query
 	) {
 		return httpPost(
 			endpoint,
 			service,
 			body,
+			query,
 			socketParams.connectTimeoutMs,
 			socketParams.socketTimeoutMs
 		);
 	}
 
-	/** Send a POST request with a JSON body using given timeout values. */
+	/** Send a POST request with a JSON body and query params using given timeout values. */
 	private HttpResponse<String> httpPost(
 		String endpoint,
 		String service,
 		Object body,
+		Map<String, Object> query,
 		int connectTimeoutMs,
 		int socketTimeoutMs
 	) {
 		String url = makeServiceUrl(endpoint, service);
-		HttpRequestWithBody req = Unirest.post(url)
-			.header("accept", "application/json")
-			.connectTimeout(connectTimeoutMs)
-			.socketTimeout(socketTimeoutMs);
+		HttpRequestWithBody req;
+		if (query == null || query.isEmpty()) {
+			req = Unirest.post(url)
+				.header("accept", "application/json")
+				.connectTimeout(connectTimeoutMs)
+				.socketTimeout(socketTimeoutMs);
+		} else {
+			req = Unirest.post(url)
+				.header("accept", "application/json")
+				.queryString(query)
+				.connectTimeout(connectTimeoutMs)
+				.socketTimeout(socketTimeoutMs);
+		}
 		if (usePublicEndpoints) {
 			if (accessToken != null) {
-				req.header("Authorization", "Bearer " + accessToken);
+				req.header(
+					"Authorization",
+					"Bearer " + accessToken.access_token
+				);
 			}
 		} else {
 			req
@@ -454,6 +571,7 @@ public class UCentralClient {
 			String.format("device/%s/wifiscan", serialNumber),
 			OWGW_SERVICE,
 			req,
+			null,
 			socketParams.connectTimeoutMs,
 			socketParams.wifiScanTimeoutMs
 		);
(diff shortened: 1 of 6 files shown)


src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
```diff
@@ -339,37 +441,52 @@ private HttpResponse<String> httpGet(
...
+				.header("accept", "application/json")
+				.connectTimeout(connectTimeoutMs)
+				.socketTimeout(socketTimeoutMs);
+		} else {
+			req = Unirest.post(url)
+				.header("accept", "application/json")
+				.queryString(query)
+				.connectTimeout(connectTimeoutMs)
+				.socketTimeout(socketTimeoutMs);
+		}
```
jwoos, line 483:
```suggestion
		HttpRequestWithBody req = Unirest.post(url)
				.header("accept", "application/json")
				.connectTimeout(connectTimeoutMs)
				.socketTimeout(socketTimeoutMs);
		if (query != null && !query.isEmpty()) {
			req.queryString(query);
		}
```

the above might be cleaner

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
```diff
@@ -195,27 +202,117 @@ public boolean login() {
...
+	 * POST a WebTokenRefreshRequest to refresh the access token.
+	 *
+	 * @return valid access token if success, otherwise return null.
+	 */
+	private WebTokenResult refreshToken() {
+		WebTokenRefreshRequest refreshRequest = new WebTokenRefreshRequest();
+		refreshRequest.userId = username;
+		refreshRequest.refreshToken = accessToken.refresh_token;
+		logger.debug("refresh token: {}", accessToken.refresh_token);
+		Map<String, Object> query = new HashMap<>();
```
jwoos, line 287: `Collections.singletonMap`?

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
```diff
@@ -195,27 +202,117 @@ public boolean login() {
...
+				response.getStatus(),
+				response.getBody()
+			);
+			return null;
+		}
+		try {
+			return gson.fromJson(response.getBody(), WebTokenResult.class);
+		} catch (JsonSyntaxException e) {
+			logger.error(
+				"Failed to serialize responseBody: Unexpected response:",
```
jwoos, line 308: specify WebTokenResult instead of responseBody

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
```diff
@@ -195,27 +202,117 @@ public boolean login() {
...
 		return loadSystemEndpoints();
 	}
 
+	/**
+	 * Check if an access token is expired.
+	 *
+	 * @return true if the access token is expired
+	 */
+	private boolean isAccessTokenExpired() {
+		return accessToken.created + accessToken.idle_timeout <
```
jwoos, line 233: what does idle timeout mean here? there's `expires_in` - why not use that?

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
```diff
@@ -195,27 +202,117 @@ public boolean login() {
...
+
+	/**
+	 * POST a WebTokenRefreshRequest to refresh the access token.
+	 *
+	 * @return valid access token if success, otherwise return null.
+	 */
+	private WebTokenResult refreshToken() {
+		WebTokenRefreshRequest refreshRequest = new WebTokenRefreshRequest();
+		refreshRequest.userId = username;
+		refreshRequest.refreshToken = accessToken.refresh_token;
```
jwoos, line 285: just trying to understand, does this mean that the refresh token is only used to refresh an existing access token?

jwoos (changes requested):

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
```diff
@@ -137,7 +140,10 @@ public static void verifySsl(boolean enable) {
 	 * The access token obtained from uCentralSec, needed only when using public
 	 * endpoints.
 	 */
-	private String accessToken;
+	private WebTokenResult accessToken;
+
+	/** Time window to refresh token in seconds. */
+	private int TOKEN_REFRESH_WINDOW_S = 21600;
```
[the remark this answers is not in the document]
zhiqiand: It's only documented in the task description [T127668163](https://www.internalfb.com/intern/tasks/?t=127668163)

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
zhiqiand:
```
expires_in":2592000,"idle_timeout":300
```
This is printed by access token. I think `idle_timeout` makes more sense here. But open to any suggestion

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
zhiqiand: Yes

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
```diff
@@ -195,27 +202,117 @@ public boolean login() {
...
 		// Load system endpoints
 		return loadSystemEndpoints();
 	}
 
+	/**
+	 * Check if an access token is expired.
+	 *
+	 * @return true if the access token is expired
+	 */
+	private boolean isAccessTokenExpired() {
```
elludraon, line 232: here and many other places, you should check if `accessToken == null` to avoid crashing in some edge cases

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
```diff
@@ -339,26 +441,29 @@ private HttpResponse<String> httpGet(
...
 	}
 
-	/** Send a POST request with a JSON body. */
+	/** Send a POST request with a JSON body and query params. */
 	private HttpResponse<String> httpPost(
 		String endpoint,
 		String service,
-		Object body
+		Object body,
+		Map<String, Object> query
```
elludraon, line 449: call this `parameters` like in `httpGet()` (be consistent)

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
```diff
@@ -270,6 +367,7 @@ public boolean isInitialized() {
 		) {
 			return false;
 		}
+		accessToken = getAccessToken();
```
elludraon, line 370: The important part of this diff is where you are calling `getAccessToken()`. This seems like a pretty random spot, and even though it probably works, it is totally unexpected for an "is initialized?" method to do this.

Was your intention to check expiry at the start of each thread's event loop? That approach seems okay, but it should just be a separate/explicit function call. The alternative is replacing *all* direct usage of `accessToken` with this function, but that is slightly more complicated and possibly introduces strange timing/delays.

Other problems with this:
- Access token is only used if `usePublicEndpoints` is true, so don't call this unless necessary
- This is not thread safe - in fact it is very likely that two threads will try to refresh the token at the same time. You need to use some form of synchronization to ensure that only 1 refresh is attempted at a time (and in an efficient way that doesn't involve blocking other threads unnecessarily)

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
```diff
@@ -284,6 +382,7 @@ public boolean isProvInitialized() {
 		if (!serviceEndpoints.containsKey(OWPROV_SERVICE)) {
 			return false;
 		}
+		accessToken = getAccessToken();
```
elludraon, line 385: see above

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
```diff
@@ -137,7 +140,10 @@ public static void verifySsl(boolean enable) {
 	 * The access token obtained from uCentralSec, needed only when using public
 	 * endpoints.
 	 */
-	private String accessToken;
+	private WebTokenResult accessToken;
+
+	/** Time window to refresh token in seconds. */
+	private final int TOKEN_REFRESH_WINDOW_S = 21600;
```
elludraon, line 146: Usually you should add a comment so people don't need to do the math:
```
private final int TOKEN_REFRESH_WINDOW_S = 21600; // 6 hours
```

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
```diff
@@ -137,7 +140,10 @@ public static void verifySsl(boolean enable) {
 	 * The access token obtained from uCentralSec, needed only when using public
 	 * endpoints.
 	 */
-	private String accessToken;
+	private WebTokenResult accessToken;
+
+	/** Time window to refresh token in seconds. */
```
elludraon, line 145: Time window *before expiry* ?

elludraon (changes requested):

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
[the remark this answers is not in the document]
jwoos: I guess what I want to know is if owsec agress with that view or if it's something that's left up to the microservices. I would think the latter doesn't make sense - so that 6 hour seems a bit arbitrary

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
jwoos: What I imagine "idle_timeout" to mean is that if there's no use of that token within that time period then the token is considered expired. However, if we take it to mean that this code is wrong since it never actually keeps track of last access time and uses a constant time of (creation + idle timeout).

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
zhiqiand: Yes the intention is that as `isInitialized` is called at the start of some events, checking the expiry at every implementation running. What do you mean by a separate function call? Can I make it as a function and call it at each thread's event loop as where `isInitialized` is called except putting it directly in `isInitialized`?

For the alternative, I have replaced usage of `accessToken` with `getAccessToken()` except login and somewhere it would be falling into infinite loop(For example, `httpPost` when sending token refresh request). Do you think it's a good idea to check the expiry In every httpGet/httpPost? 

It makes more sense to me checking expiry in every httpGet/httpPost, can we hold the both to ensure as long as the access token expired, it is refreshed immediately?

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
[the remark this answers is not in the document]
elludraon: Yup it's arbitrary. This should be moved into `RRMConfig`

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
elludraon: Your approach is better I think. What I mean is you should write another function like `refreshAccessToken()` to call *after* the `isInitialized()` loop because a function like "is initialized" implies you aren't doing any actions like this.

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
jwoos: I think idle_timeout is important, i just don't know what it actually means in this context

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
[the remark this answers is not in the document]
jwoos: seconded, if it's arbitrary it should be configurable

jwoos:
I actually think there's a conflation of expired tokens and tokens requiring refresh.

I think an expired token is when a token is completely invalid and requires a new login. A refresh has to occur after a configured `x` amount of time (i think that's `idle_timeout`, not some value we set ourselves). However, it seems like refresh workflows don't really do much in practice and don't even seem to be really defined in so we may want to clarify the behavior with someone

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
```diff
@@ -195,27 +202,125 @@ public boolean login() {
...
+	/**
+	 * Check if an access token is expired.
+	 *
+	 * @return true if the access token is expired
+	 */
+	private boolean isAccessTokenExpired() {
+		if(accessToken == null){
+			return true;
+		}
+		return accessToken.created + TOKEN_REFRESH_WINDOW_S <
```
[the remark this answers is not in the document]
zhiqiand: You are right. There are two conception. One is access_token is expired and we need to refresh it. Another is that the token is completely expired even for a refresh request. (I call it a refresh token expired)

In the first version of this PR. There are  two methods, one is `isAccessTokenExpired`, which uses `idle_timeout` to check if an access token is expired, another is `isRefreshTokenExpired`, which uses `TOKEN_REFRESH_WINDOW_S ` to check if this token is completely expired.

The second version(the current version) of this PR. it only has `isAccessTokenExpired` and uses `TOKEN_REFRESH_WINDOW_S ` to check if an access token is expired. `access_token == null` to check if the access token is completely expired or not.

I guess you votes for `idle_timeout` and `expired_in` to check these two conditions?  Then config `TOKEN_REFRESH_WINDOW_S` seems not meaningful. I might also have some misunderstanding about the task description. What's the role of `TOKEN_REFRESH_WINDOW_S`? @elludraon 

> if owsec says a token is expired (e.g. through expires_in) and we consider it not expired then that would be a security concern. 

BTW, `expired_in` is a number which is far larger than `TOKEN_REFRESH_WINDOW_S`. So all the operations we have are in the scope of security I think.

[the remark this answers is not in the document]
jwoos: Yes I think we should use both - I actually think `idle_timeout` is what Jeffrey meant to have.

So the flow would look like this:

- Is there a token?
  - No: Login
  - Yes: go to next step
- is the token expired? (i.e. `now >= token.created + token.expires_in`)
  - No: Go to next step
  - Yes: login
- does the token need to be refreshed? (i.e. `now > last_access + token.idle_timeout`)
  - No: just use it!
  - Yes: refresh it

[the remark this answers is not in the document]
elludraon: Can you check if "idle_timeout" is actually implemented/enforced? I originally overlooked that field but I'm also not sure it's really used. If it is *not* used, then the purpose of `TOKEN_REFRESH_WINDOW_S` was some arbitrary time to refresh a token before it did expire (so that we can use the "refresh" flow at all, instead of completely re-login).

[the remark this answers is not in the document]
jwoos: I went through the code last week and it's not used atm (refreshing a token isn't a thing across the services right now) but I imagine it WILL be used and if so, we should follow it. Maybe good to clarify with someone regarding that.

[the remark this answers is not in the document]
zhiqiand: Then it seems that there is not need to configure `TOKEN_REFRESH_WINDOW_S`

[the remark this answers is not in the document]
elludraon: Ah, I see the problem with the original task. I did not notice that, when refreshed, the new token contains the *same* `created` time so that makes `TOKEN_REFRESH_WINDOW_S` useless. So we only need logic to handle:
- `idle_timeout` (shouldn't really happen, but might as well since the code is already done)
- relogin flow upon expiry

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
```diff
@@ -195,27 +199,142 @@ public boolean login() {
...
 			return false;
 		}
-		if (!respBody.has("access_token")) {
+		if (
+			token == null || token.access_token == null ||
+				token.access_token.isEmpty()
+		) {
 			logger.error("Login failed: Missing access token");
-			logger.debug("Response body: {}", respBody.toString());
+			logger.debug("Response body: {}", response.toString());
```
elludraon, line 215: Does this work? I think you still need `response.getBody()`. Most Java classes are not going to override `toString()` and the default just prints the memory address.

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
```diff
@@ -195,27 +199,142 @@ public boolean login() {
...
-		logger.debug("Access token: {}", accessToken);
+		logger.debug("Access token: {}", accessToken.access_token);
+		logger.debug("Refresh token: {}", accessToken.refresh_token);
 
 		// Load system endpoints
 		return loadSystemEndpoints();
 	}
 
+	/**
+	 * when use public endpoints, refresh the access token if it's expired.
```
elludraon, line 228: nit: "When using"

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
```diff
@@ -256,6 +375,7 @@ private boolean loadSystemEndpoints() {
 			logger.debug("Response body: {}", respBody.toString());
 			return false;
 		}
+		refreshAccessToken();
```
elludraon, line 378: Why is this here?

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
```diff
@@ -195,27 +199,142 @@ public boolean login() {
...
+		}
+	}
+
+	/**
+	 * Check if the token is completely expired even if
+	 * for a token refresh request
+	 *
+	 * @return true if the refresh token is expired
+	 */
+	private boolean isRefreshTokenExpired() {
```
elludraon, line 242: this one should be named "isAccessTokenExpired"

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
```diff
@@ -195,27 +199,142 @@ public boolean login() {
...
+	 *
+	 * @return true if the access token is expired
+	 */
+	private boolean isAccessTokenExpired() {
+		if (accessToken == null) {
+			return true;
+		}
+		return accessToken.created + accessToken.idle_timeout <
+			Instant.now().getEpochSecond();
+	}
```
elludraon, line 261: This is wrong. You are checking if the access token has *timed out*, not expired (so name it "isAccessTokenTimedOut" or similar). Also, you need to keep track of the last use time, not `created`.

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
```diff
@@ -195,27 +199,142 @@ public boolean login() {
...
+			return null;
+		}
+		if (isRefreshTokenExpired()) {
+			logger.info("Token is expired, login again");
+			if (login()) {
+				return accessToken;
+			}
+			return null;
+		} else if (isAccessTokenExpired()) {
+			logger.debug("Access token is expired, start refreshing a token.");
```
elludraon, line 281: "Access token timed out, refreshing the token"

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
```diff
@@ -195,27 +199,142 @@ public boolean login() {
...
 		// Load system endpoints
 		return loadSystemEndpoints();
 	}
 
+	/**
+	 * when use public endpoints, refresh the access token if it's expired.
+	 */
+	public synchronized void refreshAccessToken() {
+		if (usePublicEndpoints) {
+			accessToken = getAccessToken();
```
elludraon, line 232: nit: I would prefer if you don't always set this field esp. if it didn't change. You can also rename the sub-function something clearer like `refreshAccessTokenImpl()` (for "implementation"), make it return `void`, and set the token there only when needed.

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
```diff
@@ -195,27 +199,142 @@ public boolean login() {
...
+	}
+
+	/**
+	 * Check if an access token is expired.
+	 *
+	 * @return true if the access token is expired
+	 */
+	private boolean isAccessTokenExpired() {
+		if (accessToken == null) {
+			return true;
```
elludraon, line 257: This should be false? Or, make sure you are not trying to refresh a token when it is null.

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
```diff
@@ -195,27 +199,142 @@ public boolean login() {
...
+		logger.debug("Refresh token: {}", accessToken.refresh_token);
 
 		// Load system endpoints
 		return loadSystemEndpoints();
 	}
 
+	/**
+	 * when use public endpoints, refresh the access token if it's expired.
+	 */
+	public synchronized void refreshAccessToken() {
```
elludraon, line 230: While this `synchronized` *works*, this is not a good location because it will very frequently trigger locking for no reason. You only want to lock when necessary, i.e. refreshing/re-login.

Here's one possible/simple way to only lock when needed:
```java
private WebTokenResult getAccessToken() {
	// ...
	if (isRefreshTokenExpired()) {
		synchronized (this) {			
			if (isRefreshTokenExpired()) {
				// inner logic here
			}
		}
	} else if (isAccessTokenExpired()) {
		synchronized (this) {
			if (isAccessTokenExpired()) {
				// inner logic here
			}
		}
	}
}
```
In the "normal case", no locking is done. When the token is actually expired, then all threads will block while the first thread obtains a new token, then other threads unblock and see that the token is no longer expired. Do take care to handle error cases (if first thread fails and token is now null).

elludraon (changes requested):

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
zhiqiand: It should be false to indicate that the access token is invalid and considered as expired/timeout

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
```diff
@@ -195,27 +199,142 @@ public boolean login() {
...
+	/**
+	 * Check if an access token is expired.
+	 *
+	 * @return true if the access token is expired
+	 */
+	private boolean isAccessTokenTimedOut() {
+		if (accessToken == null) {
+			return true;
+		}
+		return accessToken.created + accessToken.idle_timeout <
```
jwoos, line 259: I think we need to keep track of last accessed time and then use `last_accessed + accessToken.idle_timeout` since idle timeout only starts applying from last access

jwoos (commented):
I know this has been dragging on for a while but one last comment from my end re: how we calculate timeout

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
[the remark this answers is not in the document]
elludraon: Update: It appears that the "refreshToken" call is returning "created" time of the LAST token, not the current/refreshed one. So that is the confusion here. 😢

jwoos (approved):
Thanks for addressing everything!

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
```diff
@@ -142,6 +142,17 @@ public static void verifySsl(boolean enable) {
...
+	 * The unix timestamp (in seconds) keeps track of when the accessToken is created.
+	 */
+	private long created;
+
+	/**
+	 * The unix timestamp (in seconds) keeps track of last time when the accessToken 
+	 * is accessed.
+	 */
+	private long lastAccess;
+
```
elludraon, line 155: can you pick better names like "accessTokenCreated" and "accessTokenLastAccess"?

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
```diff
@@ -280,6 +293,8 @@ private void refreshAccessTokenImpl() {
 				if (isAccessTokenTimedOut()) {
 					logger.debug("Access token timed out, refreshing the token");
 					accessToken = refreshToken();
+					created = Instant.now().getEpochSecond();
+					lastAccess = created;
```
elludraon, line 297: can you update this in the `refreshToken()` function instead?

src/main/java/com/facebook/openwifirrm/ucentral/UCentralClient.java
```diff
@@ -142,6 +142,17 @@ public static void verifySsl(boolean enable) {
 	 */
 	private WebTokenResult accessToken;
 
+	/**
+	 * The unix timestamp (in seconds) keeps track of when the accessToken is created.
+	 */
+	private long created;
```
elludraon, line 148: Did you confirm this is needed?? (and shouldn't be fixed in owsec?)

zhiqiand: Stephan didn't reply to me yet...

PR #89, opened 2022-09-28
[WIFI-10819] parse cron into valid quartz cron

jwoos:
Linux cron is weird and ORs the day of month and day of week values are both not `*`. Quartz is even weirder in that it can't support crons with both values specified.

>  Note: The day of a command's execution can be specified in the
       following two fields — 'day of month', and 'day of week'.  If
       both fields are restricted (i.e., do not contain the "*"
       character), the command will be run when either field matches the
       current time.  For example,
       "30 4 1,15 * 5" would cause a command to be run at 4:30 am on the
       1st and 15th of each month, plus every Friday.

From https://man7.org/linux/man-pages/man5/crontab.5.html

Since it's simply matching either of them, split them out into two different crons and register them both. I had to change the type of a field, which may or may not cause problems depending on if there is a configuration file to read or not.

Signed-off-by: Jun Woo Shin <jwoos@fb.com>

### DIFF
--- a/src/main/java/com/facebook/openwifirrm/RRMSchedule.java
+++ b/src/main/java/com/facebook/openwifirrm/RRMSchedule.java
@@ -19,9 +19,9 @@ public class RRMSchedule {
 	 *
 	 * This field expects a cron-like format as defined by the Quartz Job
 	 * Scheduler (CronTrigger):
-	 * https://www.quartz-scheduler.org/documentation/quartz-2.3.0/tutorials/crontrigger.html
+	 * https://www.quartz-scheduler.org/documentation/quartz-2.4.0/tutorials/crontrigger.html
 	 */
-	public String cron;
+	public String[] cron;
 
 	/**
 	 * The list of RRM algorithms to run.
--- a/src/main/java/com/facebook/openwifirrm/RRMSchedule.java
+++ b/src/main/java/com/facebook/openwifirrm/RRMSchedule.java
@@ -21,7 +21,7 @@ public class RRMSchedule {
 	 * Scheduler (CronTrigger):
 	 * https://www.quartz-scheduler.org/documentation/quartz-2.4.0/tutorials/crontrigger.html
 	 */
-	public String[] cron;
+	public List<String> cron;
 
 	/**
 	 * The list of RRM algorithms to run.
--- a/src/main/java/com/facebook/openwifirrm/RRMSchedule.java
+++ b/src/main/java/com/facebook/openwifirrm/RRMSchedule.java
@@ -21,7 +21,7 @@ public class RRMSchedule {
 	 * Scheduler (CronTrigger):
 	 * https://www.quartz-scheduler.org/documentation/quartz-2.4.0/tutorials/crontrigger.html
 	 */
-	public List<String> cron;
+	public List<String> crons;
 
 	/**
 	 * The list of RRM algorithms to run.
--- a/src/main/java/com/facebook/openwifirrm/modules/ProvMonitor.java
+++ b/src/main/java/com/facebook/openwifirrm/modules/ProvMonitor.java
@@ -159,12 +159,22 @@ public class ProvMonitor implements Runnable {
 			return null;
 		}
 
-		RRMSchedule schedule = new RRMSchedule();
-		schedule.cron = RRMScheduler
+		// RRMScheduler.parseIntoQuartzCron will only return an array of length 1 or 2
+		String[] crons = RRMScheduler
 			.parseIntoQuartzCron(details.rrm.schedule);
-		if (schedule.cron == null || schedule.cron.isEmpty()) {
+		if (crons == null || crons.length == 0) {
 			return null;
 		}
+		// if ANY crons are invalid throw it out since it doesn't make sense to
+		// schedule partial jobs
+		for (String cron : crons) {
+			if (cron == null || cron.isEmpty()) {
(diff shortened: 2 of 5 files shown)


src/main/java/com/facebook/openwifirrm/modules/ProvMonitor.java
```diff
@@ -159,12 +159,22 @@ protected RRMSchedule transformDetailsToSchedule(RRMDetails details) {
 			return null;
 		}
 
-		RRMSchedule schedule = new RRMSchedule();
-		schedule.cron = RRMScheduler
+		// RRMScheduler.parseIntoQuartzCron will only return an array of length 1 or 2
```
elludraon, line 162: this comment seems unnecessary (not important to logic, maybe will change?)

src/main/java/com/facebook/openwifirrm/modules/RRMScheduler.java
```diff
@@ -125,7 +134,8 @@ public static String parseIntoQuartzCron(String linuxCron) {
 		final int DAY_OF_MONTH_INDEX = 3;
 		final int DAY_OF_WEEK_INDEX = 5;
 
-		String dayOfMonth = split[DAY_OF_MONTH_INDEX];
+		// make a copy since it's possible it'll be used below
+		String dayOfMonth = new String(split[DAY_OF_MONTH_INDEX]);
```
elludraon, line 138: I don't understand why you're copying - Java String is immutable

src/main/java/com/facebook/openwifirrm/modules/RRMScheduler.java
```diff
@@ -144,15 +154,32 @@ public static String parseIntoQuartzCron(String linuxCron) {
...
+
+			split[DAY_OF_MONTH_INDEX] = dayOfMonth;
+			split[DAY_OF_WEEK_INDEX] = "?";
+			String dayOfMonthCron = String.join(" ", split);
+
+			if (
+				!CronExpression.isValidExpression(dayOfWeekCron) ||
+					!CronExpression.isValidExpression(dayOfMonthCron)
+			) {
+				return null;
```
elludraon, line 171: log an error here?

src/main/java/com/facebook/openwifirrm/modules/RRMScheduler.java
```diff
@@ -74,15 +75,22 @@ public class RRMScheduler {
 	/** The scheduler instance. */
 	private Scheduler scheduler;
 
-	/** The zones with active triggers scheduled. */
-	private Set<String> scheduledZones;
+	/**
+	 * The job keys with active triggers scheduled. Job keys take the format of
+	 * <zone>:<index>, where index applies only if there are multiple schedules for
+	 * one zone
```
elludraon, line 81: * no need for the text about `index` - it's a bit misleading anyway
* pretty sure `<zone>:<index>` will render incorrectly because `<>` are getting parsed as html tags - pick a different character or put it in a code block

src/main/java/com/facebook/openwifirrm/modules/RRMScheduler.java
```diff
@@ -74,15 +75,22 @@ public class RRMScheduler {
...
+	 *
+	 * @see RRMScheduler.parseIntoQuartzCron
```
elludraon, line 83: is this formatted correctly? I thought it should be:
```
@see #parseIntoQuartzCron(String)
```

src/main/java/com/facebook/openwifirrm/modules/RRMScheduler.java
```diff
@@ -218,85 +245,105 @@ public void shutdown() {
...
-						"Invalid cron expression (%s) for zone %s",
-						config.schedule.cron,
-						zone
-					),
-					e
-				);
+			if (
+				schedule.cron == null ||
+					schedule.cron.length == 0
+			) {
```
elludraon, line 271: duplicated w/ above?

src/main/java/com/facebook/openwifirrm/modules/RRMScheduler.java
```diff
@@ -218,85 +245,105 @@ public void shutdown() {
...
-				if (!prevScheduled.contains(zone)) {
-					scheduler.scheduleJob(trigger);
-				} else {
-					scheduler.rescheduleJob(trigger.getKey(), trigger);
+			for (int i = 0; i < schedule.cron.length; i++) {
+				String cron = schedule.cron[i];
+
+				// if even one schedule has invalid cron, the whole thing is probably wrong
+				if (cron == null || cron.isEmpty()) {
+					break;
```
elludraon, line 280: log error

src/main/java/com/facebook/openwifirrm/RRMSchedule.java
```diff
@@ -21,7 +21,7 @@ public class RRMSchedule {
 	 * Scheduler (CronTrigger):
 	 * https://www.quartz-scheduler.org/documentation/quartz-2.4.0/tutorials/crontrigger.html
 	 */
-	public String[] cron;
+	public List<String> cron;
```
elludraon, line 24: Oh, forgot to write this last time - if you rename this to `crons` it will avoid any compatibility issues/crashing (and more appropriate name anyway).

elludraon (approved):
Looks good! (one last comment)

PR #90, opened 2022-09-29
Refactor IE parsing

RockyMandayam2:
- Use InformationElements wrapper class
- Catch Exception for parsing errors instead of NullPointerException
- Add/edit javadocs

### DIFF
--- a/src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
+++ b/src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
@@ -120,23 +120,28 @@ public class UCentralUtils {
 			}
 			JsonObject contents =
 				ie.getAsJsonObject().get(IE_CONTENT_FIELD_KEY).getAsJsonObject();
-			switch (typeElement.getAsInt()) {
-			case Country.TYPE:
-				ies.country = Country.parse(contents);
-				break;
-			case QbssLoad.TYPE:
-				ies.qbssLoad = QbssLoad.parse(contents);
-				break;
-			case LocalPowerConstraint.TYPE:
-				ies.localPowerConstraint =
-					LocalPowerConstraint.parse(contents);
-				break;
-			case TxPwrInfo.TYPE:
-				ies.txPwrInfo = TxPwrInfo.parse(contents);
-				break;
+			try {
+				switch (typeElement.getAsInt()) {
+				case Country.TYPE:
+					ies.country = Country.parse(contents);
+					break;
+				case QbssLoad.TYPE:
+					ies.qbssLoad = QbssLoad.parse(contents);
+					break;
+				case LocalPowerConstraint.TYPE:
+					ies.localPowerConstraint =
+						LocalPowerConstraint.parse(contents);
+					break;
+				case TxPwrInfo.TYPE:
+					ies.txPwrInfo = TxPwrInfo.parse(contents);
+					break;
+				}
+			} catch (Exception e) {
+				logger.debug("Skipping invalid IE.", e);
+				continue;
 			}
 		}
-		entry.ies = ies;
+		entry.ieContainer = ies;
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


src/main/java/com/facebook/openwifirrm/ucentral/WifiScanEntry.java
```diff
@@ -38,23 +31,14 @@ public WifiScanEntry() {}
 	public WifiScanEntry(WifiScanEntry o) {
 		super(o);
 		this.unixTimeMs = o.unixTimeMs;
-		this.country = o.country;
-		this.qbssLoad = o.qbssLoad;
-		this.localPowerConstraint = o.localPowerConstraint;
-		this.txPwrInfo = o.txPwrInfo;
+		this.ies = o.ies; // TODO deep copy?
```
RockyMandayam2, line 34: Should I deep copy here? Either way, I'll address and delete this TODO comment

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
```diff
@@ -111,34 +111,32 @@ private static void extractIEs(
...
-					entry.localPowerConstraint =
-						LocalPowerConstraint.parse(contents);
-					break;
-				case TxPwrInfo.TYPE:
-					entry.txPwrInfo = TxPwrInfo.parse(contents);
-					break;
-				}
-			} catch (NullPointerException e) {
-				logger.debug("Skipping invalid IE.", e);
+			switch (typeElement.getAsInt()) {
```
RockyMandayam2, line 123: I removed the `NullPointerException` check here, but I'm not sure the best way to proceed... We don't want a single invalid IE to crash the `Modeler` thread, right? But doing a null check for every single field instead of catching an exception (which shouldn't happen often) seems like too much extra code? Thoughts?

Based on this, I may change all the IEs. Currently, mandatory fields are assumed to exist in all the IEs...

src/main/java/com/facebook/openwifirrm/ucentral/informationelement/Country.java
```diff
@@ -124,15 +125,8 @@ public static Country parse(JsonObject contents) {
...
-				try {
-					constraints
-						.add(CountryInfo.parse(jsonElement.getAsJsonObject()));
-				} catch (NullPointerException e) {
-					logger.debug(
-						"Skipping invalid constraint encountered in Country IE.",
-						e
-					);
-				}
+				CountryInfo countryInfo = CountryInfo.parse(jsonElement.getAsJsonObject());
```
RockyMandayam2, line 128: Similar to my other comment - a null check for every field instead of catching an exception (which shouldn't happen often) seems like too much extra code? Thoughts?

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
RockyMandayam2: Also, keep in mind that the plan is to also parse in maybe another 10 IEs...

src/main/java/com/facebook/openwifirrm/ucentral/WifiScanEntry.java
```diff
@@ -26,10 +22,7 @@ public class WifiScanEntry extends WifiScanEntryResult {
 	 * time reference.
 	 */
 	public long unixTimeMs;
-	public Country country;
-	public QbssLoad qbssLoad;
-	public LocalPowerConstraint localPowerConstraint;
-	public TxPwrInfo txPwrInfo;
+	public InformationElements ies;
```
elludraon, line 25: - don't use the same name `ies`, it will be confusing -- `ie` or `ieContainer` or such?
- add javadoc

src/main/java/com/facebook/openwifirrm/ucentral/WifiScanEntry.java
elludraon: I think shallow is fine, we're not planning to modify any fields

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
elludraon: If you really feel it is too much code, you can `catch (Exception e)` here (not specifically `NullPointerException`)

elludraon (approved):

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
RockyMandayam2: I will catch `Exception`

PR #91, opened 2022-09-29
Rename HTOperationElement, HTOperationElementTest, VHTOperationElement, and VHTOperationElementTest

RockyMandayam2:
Rename HTOperationElement, HTOperationElementTest, VHTOperationElement, and VHTOperationElementTest for consistency (all other classes in the package do not include the "Element" suffix)


zhiqiand (approved):
LGTM. Seems that only class names are changed.

PR #92, opened 2022-09-30
[WIFI-10943] Deal with "auto" value for channel and fix 80p80 representation

jwoos:
1. "auto" is a valid value in configuration - there was an assumption that it was only an int and it tried to call `getAsInt()` without any safeguards. This caused the exception to propagate up (not caught anywhere) and crash the system.
    1. https://github.com/Telecominfraproject/wlan-ucentral-schema/blob/bfcb0e8c6e6c07195ba964f696c562aae3723fa3/schema/radio.yml#L27-L36
3. 80p80 was represented as 8080, which is problematic as the value is taken at face value i.e. it's used directly in computations. So it was seen has 8080MHz instead of the proper 80p80 it is. @pohanhf can speak more to this.

### DIFF
--- a/src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
+++ b/src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
@@ -193,8 +193,8 @@ public class UCentralUtils {
 			) {
 				currentValue = fieldValue.getAsInt();
 			} else {
-				logger.error(
-					"Unable to get field {} as int, value was {}",
+				logger.debug(
+					"Unable to get field '{}' as int, value was {}",
 					fieldName,
 					fieldValue.toString()
 				);
--- a/src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
+++ b/src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
@@ -187,18 +187,19 @@ public class UCentralUtils {
 			// not all values are int so override those values
 			Integer currentValue = null;
 			JsonElement fieldValue = radioConfig.get(fieldName);
-			try {
+			if (
+				fieldValue.isJsonPrimitive() &&
+					fieldValue.getAsJsonPrimitive().isNumber()
+			) {
 				currentValue = fieldValue.getAsInt();
-			} catch (NumberFormatException e) {
+			} else {
 				logger.error(
-					String.format(
-						"Unable to get field %s as int, value was %s",
-						fieldName,
-						fieldValue.toString()
-					),
-					e
+					"Unable to get field {} as int, value was {}",
+					fieldName,
+					fieldValue.toString()
 				);
 			}
+
 			if (currentValue != null && currentValue == newValue) {
 				logger.info(
 					"Device {}: {} {} is already {}",
--- a/src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
+++ b/src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
@@ -183,9 +183,23 @@ public class UCentralUtils {
 				continue;
 			}
 
-			// Compare vs. existing value
-			int currentValue = radioConfig.get(fieldName).getAsInt();
-			if (currentValue == newValue) {
+			// Compare vs. existing value.
+			// not all values are int so override those values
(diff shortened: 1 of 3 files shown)


src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
```diff
@@ -183,9 +183,23 @@ public static boolean setRadioConfigField(
...
-			// Compare vs. existing value
-			int currentValue = radioConfig.get(fieldName).getAsInt();
-			if (currentValue == newValue) {
+			// Compare vs. existing value.
+			// not all values are int so override those values
+			Integer currentValue = null;
+			JsonElement fieldValue = radioConfig.get(fieldName);
+			try {
+				currentValue = fieldValue.getAsInt();
+			} catch (NumberFormatException e) {
```
elludraon, line 192: Instead of doing this via exception handling you could write:
```
JsonElement e;
if (e.isJsonPrimitive() && e.getAsJsonPrimitive().isNumber()) {
  e.getAsInt();
}
```
These JSON libraries don't make it easy... but it's possible.

elludraon (approved):

src/test/java/com/facebook/openwifirrm/ucentral/UCentralUtilsTest.java
```diff
@@ -17,4 +22,60 @@ public class UCentralUtilsTest {
 	void test_placeholder() throws Exception {
 		assertEquals(3, 1 + 2);
 	}
+
+	@Test
+	void test_setRadioConfigFieldChannel() throws Exception {
+		final String serialNumber = "aaaaaaaaaaaa";
+		final int expectedChannel = 10;
```
RockyMandayam2, line 29: Minor, maybe just a preference: maybe add `  // arbitrary` to the right of `10;`? Only because `10` is not a commonly used channel. Or change it to `1`?

RockyMandayam2 (changes requested):

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
jwoos: Yeah I thought about it but then I went back to this but I have no real preference so let me do that

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
```diff
@@ -183,9 +183,24 @@ public static boolean setRadioConfigField(
...
+			// not all values are int so override those values
+			Integer currentValue = null;
+			JsonElement fieldValue = radioConfig.get(fieldName);
+			if (
+				fieldValue.isJsonPrimitive() &&
+					fieldValue.getAsJsonPrimitive().isNumber()
+			) {
+				currentValue = fieldValue.getAsInt();
+			} else {
+				logger.error(
```
RockyMandayam2, line 196: This should be `logger.debug` since getting "auto" is not an error.

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
```diff
@@ -183,9 +183,24 @@ public static boolean setRadioConfigField(
...
+					"Unable to get field {} as int, value was {}",
```
RockyMandayam2, line 197: Maybe put quotes around field name? Like `field '{}'`

RockyMandayam2 (changes requested):

pohanhf:
For the 8080 one, we right now downgrade it to the 20MHz case here in https://github.com/Telecominfraproject/wlan-cloud-rrm/blob/main/src/main/java/com/facebook/openwifirrm/optimizers/channel/ChannelOptimizer.java#L179-L184 and https://github.com/Telecominfraproject/wlan-cloud-rrm/blob/main/src/main/java/com/facebook/openwifirrm/optimizers/channel/ChannelOptimizer.java#L258-L261. 

However, we actually can determine the first set of 80MHz. We return 80MHz for now for that set for consistency with the assignment part. We will revisit this and add the second set of 80MHz properly.

Thanks Jun Woo for the workaround. :)

pohanhf (approved):

RockyMandayam2 (approved):

PR #94, opened 2022-10-03
Handle invalid IEs

RockyMandayam2:
Handle IEs (before the try-catch):
- without a TYPE field
- that aren't (json) objects
- without a contents field

Also, instead of printing the exception in the log debug message, print the IE that was skipped.

### DIFF
--- a/src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
+++ b/src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
@@ -124,12 +124,11 @@ public class UCentralUtils {
 				continue;
 			}
 			JsonObject ie = ieJsonElement.getAsJsonObject();
-			if (!ie.has(IE_CONTENT_FIELD_KEY)) {
-				// the IEs we are interested in have the content field
-				continue;
-			}
 			JsonElement contentsJsonElement = ie.get(IE_CONTENT_FIELD_KEY);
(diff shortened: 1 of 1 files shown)


src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
```diff
@@ -113,13 +113,26 @@ private static void extractIEs(
...
+			if (!ieJsonElement.isJsonObject()) {
+				// the IEs we are interested in are Json objects
+				continue;
+			}
+			JsonObject ie = ieJsonElement.getAsJsonObject();
+			if (!ie.has(IE_CONTENT_FIELD_KEY)) {
+				// the IEs we are interested in have the content field
+				continue;
+			}
+			JsonElement contentsJsonElement = ie.get(IE_CONTENT_FIELD_KEY);
```
jwoos, line 131: nit: we can just call get and then check if it's null to see if the field exists

jwoos (approved):
thanks, looks good

src/main/java/com/facebook/openwifirrm/ucentral/UCentralUtils.java
RockyMandayam2: Yes, I should definitely do that.

elludraon (approved):

PR #96, opened 2022-10-06
Follow-up on stats aggregation

zhiqiand:

This PR should resolve some conversation from Jerrey in https://github.com/Telecominfraproject/wlan-cloud-rrm/pull/82

- Didn't use `sychronizedList` as it is thread-safe already. 

> In RRM algorithms, the whole DataModel is copied.

- Didn't override `hashCode()` in `AggregatedState`, use a check method instead. Remove `hashCode()` and `equals()`
-  use `while` loop to truncate the State buffer to the correct size.

### DIFF
--- a/src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
+++ b/src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
@@ -170,11 +170,6 @@ public class AggregatedState {
 	}
 
 	@Override
-	public int hashCode() {
-		return Objects.hash(bssid, station, radio);
-	}
-
-	@Override
 	public boolean equals(Object obj) {
 		if (this == obj) {
 			return true;
@@ -202,15 +197,27 @@ public class AggregatedState {
 	}
 
 	/**
+	 * Check whether some other AggregatedState and this one are matached for aggregation.
+	 * If the two match in bssid, station and radio. Then they could be aggregated to one.
+	 *
+	 * @param state
+	 * @return boolean
+	 */
+	public boolean ifMatchForAggregation(AggregatedState state) {
+		return bssid == state.bssid && station == state.station &&
+			Objects.equals(radio, state.radio);
+	}
+
+	/**
 	 * Add an AggregatedState to this AggregatedState. Succeed only when the two
-	 * matches in hashCode.
+	 * match for aggregation.
 	 *
 	 * @param state input AggregatedState
-	 * @return boolean true if the two matches in bssid, station, channel,
+	 * @return boolean true if the two match in bssid, station, channel,
 	 *         channel_width and tx_power
 	 */
 	public boolean add(AggregatedState state) {
-		if (hashCode() == state.hashCode()) {
+		if (ifMatchForAggregation(state)) {
 			this.rssi.addAll(state.rssi);
 			this.rxRate.add(state.rxRate);
 			this.txRate.add(state.txRate);
--- a/src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
+++ b/src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
@@ -169,41 +169,14 @@ public class AggregatedState {
 		this.radio = new Radio(radioInfo);
 	}
 
-	@Override
-	public boolean equals(Object obj) {
-		if (this == obj) {
-			return true;
-		}
-		if (obj == null) {
-			return false;
-		}
-		if (getClass() != obj.getClass()) {
-			return false;
-		}
-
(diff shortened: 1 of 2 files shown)


src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
```diff
@@ -169,11 +169,6 @@ public AggregatedState(
 		this.radio = new Radio(radioInfo);
 	}
 
-	@Override
-	public int hashCode() {
```
RockyMandayam2, line 173: It is good practice to override `hashCode` and `equals` or to override neither. I think the solution is to implement `hashCode` (using Eclipse's automatic tool to do this) and just not use it... Alternatively, if `equals` is never used, we could remove both `hashCode` and `equals`

Jeffrey mentioned they should not differ in which variables they check as well. (Actually, I don't see a problem with `hashCode` checking _fewer_ variables than equals, since AFAIK we only must guarantee that objects that are `equal` have the same `hashCode`, and we don't care about the other direction. That being said, maybe I'm missing something, and maybe it is more confusing to have `hashCode` look at fewer variables, so I'm fine with the idea of keeping them consistent.)

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
```diff
@@ -201,16 +196,28 @@ public boolean equals(Object obj) {
...
 	}
 
+	/**
+	 * Check whether some other AggregatedState and this one are matached for aggregation.
+	 * If the two match in bssid, station and radio. Then they could be aggregated to one.
+	 *
+	 * @param state
+	 * @return boolean
+	 */
+	public boolean ifMatchForAggregation(AggregatedState state) {
```
RockyMandayam2, line 206: I like the name `matchesForAggregation` or `shouldAggregate` better

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
```diff
@@ -201,16 +196,28 @@ public boolean equals(Object obj) {
 			Objects.equals(radio, other.radio);
 	}
 
+	/**
+	 * Check whether some other AggregatedState and this one are matached for aggregation.
+	 * If the two match in bssid, station and radio. Then they could be aggregated to one.
+	 *
+	 * @param state
```
RockyMandayam2, line 203: Either add a description for this param or remove this line

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
```diff
@@ -201,16 +196,28 @@ public boolean equals(Object obj) {
...
+	 * @return boolean
```
RockyMandayam2, line 204: Either add a description for the return value or remove this line

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
```diff
@@ -201,16 +196,28 @@ public boolean equals(Object obj) {
 			Objects.equals(radio, other.radio);
 	}
 
+	/**
+	 * Check whether some other AggregatedState and this one are matached for aggregation.
```
RockyMandayam2, line 200: `Check whether some other AggregatedState and this one are matached for aggregation` -> `Check whether the passed-in AggregatedState and this one match for aggregation`

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
```diff
@@ -201,16 +196,28 @@ public boolean equals(Object obj) {
...
+	 * If the two match in bssid, station and radio. Then they could be aggregated to one.
```
RockyMandayam2, line 201: I think: `radio. Then they could be aggregated to one` -> `radio, then they should be aggregated`

RockyMandayam2 (changes requested):

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
elludraon, line 206: please rename to `matchesForAggregation` (no "if")

elludraon (approved):

src/main/java/com/facebook/openwifirrm/ucentral/models/AggregatedState.java
zhiqiand: As it's not used at this time, I will delete it.

RockyMandayam2 (approved):

PR #98, opened 2022-10-11
[WIFI-10910] Chg: internal endpoint port

dunaev-opsfleet:
Signed-off-by: Dmitry Dunaev <dmitry@opsfleet.com>


jwoos (approved):
thank you!

PR #99, opened 2022-10-12
[WIFI-11196] Restructure repository into multimodule project

elludraon:
Restructure the repository from a single-module Maven project to a multi-module Reactor project. This mainly splits the boilerplate code related to the OpenWiFi CloudSDK out of the base project.

Changes:
- Most original directory contents moved into `owrrm` subdirectory
- CloudSDK specific code moved to new subdirectory `lib-cloudsdk`
- New root (parent) directory declares submodules `owrrm`, `lib-cloudsdk` as a Reactor project
- Package `com.facebook.openwifirrm` moved to `com.facebook.openwifi.rrm`
- Subpackage `ucentral` moved under `cloudsdk` and refactored
- Javadoc build has changed from `mvn javadoc:javadoc` to `mvn javadoc:aggregate`

Minor notes:
- `UCentralUtils.channelToFrequencyMHz()` implementation is simplified to avoid dependency on `ChannelOptimizer.AVAILABLE_CHANNELS_BAND`
- `UCentralUtils.generateServiceKey()` is moved to `Utils` to avoid dependency on `RRMConfig.ServiceConfig`
- `UCentralClient` constructor parameter `UCentralSocketParams` is replaced with individual members to avoid dependency on `RRMConfig`
- `AggregatedState` is moved upwards out of the `models` subpackage
- Fixed various Javadoc errors in existing code (`ModelerUtils`)


lib-cloudsdk/pom.xml
```diff
@@ -0,0 +1,80 @@
...
+    <finalName>openwifi-cloudsdk</finalName>
+    <resources>
+      <resource>
+        <directory>src/main/resources</directory>
+        <includes>
+          <include>**/*</include>
+        </includes>
+      </resource>
+    </resources>
+    <plugins>
```
RockyMandayam2, line 25: Is it possible to / Is it recommended to take the common plugins and move them up into the higher level pom?

RockyMandayam2 (approved):
Thanks! Just left one comment.

jwoos (approved):
if it builds, it ships

lib-cloudsdk/pom.xml
elludraon: They are already in the parent pom, that's the `pluginManagement` section (and similarly `dependencyManagement`), hence why there are only stubs in the child pom.

PR #100, opened 2022-10-12
Move AVAILABLE_CHANNELS_BAND to UCentralUtils before adding 6G support

RockyMandayam2:
Move AVAILABLE_CHANNELS_BAND to UCentralUtils before adding 6G support. (Separating changes related to moving this constant from the more 6G-specific changes to come)

### DIFF
--- a/lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/UCentralUtils.java
+++ b/lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/UCentralUtils.java
@@ -61,25 +61,28 @@ public class UCentralUtils {
 	}
 
 	/** List of available channels per band for use. */
-	public static final Map<String, List<Integer>> AVAILABLE_CHANNELS_BAND =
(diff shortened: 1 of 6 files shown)


lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/UCentralUtils.java
```diff
@@ -58,6 +60,24 @@ public class UCentralUtils {
 		UCentralUtils.UPPER_CHANNEL_LIMIT.put(UCentralConstants.BAND_5G, 165);
 	}
 
+	/** List of available channels per band for use. */
+	public static final Map<String, List<Integer>> AVAILABLE_CHANNELS_BAND =
```
elludraon, line 64: I would suggest declaring this as `Collections.unmodifiableMap()` esp. if this is going into a util class. You can write some private static function for this instead of doing it in a `static` block.

elludraon (approved):

PR #104, opened 2022-10-13
Use device capabilities to identify radio band

RockyMandayam2:
This is done in anticipation of adding 6G support. 6G uses the same channel numbers as other bands. E.g., channel 1 could refer to 2G channel 1 or 6G channel 1. Therefore, we cannot just infer band form channel anymore.

This PR makes sure we use device capabilities to identify a radio's band separately from its channel number, and that when we categorize radios based on channel, we also take band into account (i.e., we should categorize on the (band, channel) tuple instead).

### DIFF
--- a/owrrm/src/main/java/com/facebook/openwifi/rrm/modules/ModelerUtils.java
+++ b/owrrm/src/main/java/com/facebook/openwifi/rrm/modules/ModelerUtils.java
@@ -30,6 +30,8 @@ import com.facebook.openwifi.cloudsdk.models.ap.State.Interface.SSID.Association
 import com.facebook.openwifi.rrm.aggregators.Aggregator;
 import com.facebook.openwifi.rrm.aggregators.MeanAggregator;
 import com.facebook.openwifi.rrm.modules.Modeler.DataModel;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
 
 /**
  * Modeler utilities.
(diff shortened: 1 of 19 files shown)


owrrm/src/main/java/com/facebook/openwifi/rrm/modules/ModelerUtils.java
```diff
@@ -555,4 +557,21 @@ public static String getBssidStationKeyPair(String bssid, String station) {
 			station
 		);
 	}
+
+	/** Return the radio's band, or null if band cannot be found */
+	public static String getBand(
+		State.Radio radio,
+		JsonObject deviceCapability
+	) {
+		JsonElement radioCapabilityElement = deviceCapability.get(radio.phy);
```
elludraon, line 566: I would do a null check on `radio.phy` too

owrrm/src/main/java/com/facebook/openwifi/rrm/modules/ModelerUtils.java
```diff
@@ -555,4 +557,21 @@ public static String getBssidStationKeyPair(String bssid, String station) {
...
+		if (radioCapabilityElement == null) {
+			return null;
+		}
+		JsonObject radioCapability = radioCapabilityElement.getAsJsonObject();
+		JsonElement bandsElement = radioCapability.get("band");
+		if (bandsElement == null) {
+			return null;
+		}
+		return bandsElement.getAsJsonArray().get(0).getAsString();
```
elludraon, line 575: also check if array is empty before `get(0)` ?

owrrm/src/main/java/com/facebook/openwifi/rrm/optimizers/tpc/MeasurementBasedApApTPC.java
```diff
@@ -217,7 +219,13 @@ protected static Map<String, List<Integer>> buildRssiMap(
 
 			// At a given AP, if we receive a signal from ap_2, then it gets added to the rssi list for ap_2
 			latestScan.stream()
-				.filter(entry -> (managedBSSIDs.contains(entry.bssid) && UCentralUtils.isChannelInBand(entry.channel, band)))
+				.filter((entry) -> {
+					boolean isManaged = managedBSSIDs.contains(entry.bssid);
```
elludraon, line 223: nit: replace with `if (!managedBSSIDs.contains(...)) return false;` - exit early instead of always running the next check

elludraon (approved):

PR #105, opened 2022-10-14
[WIFI-11227] Support 6G band

RockyMandayam2:
[WIFI-11227] Support 6G band. For now, we will only support channels that both low power and standard power APs can use.

[WIFI-11227]: https://telecominfraproject.atlassian.net/browse/WIFI-11227?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/UCentralUtils.java
+++ b/lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/UCentralUtils.java
@@ -44,29 +44,17 @@ public class UCentralUtils {
 	/** The Gson instance. */
 	private static final Gson gson = new Gson();
 
-	/** Map of band to the band-specific lowest available channel*/
-	public static final Map<String, Integer> LOWER_CHANNEL_LIMIT =
-		new HashMap<>();
-	static {
-		UCentralUtils.LOWER_CHANNEL_LIMIT.put(UCentralConstants.BAND_2G, 1);
-		UCentralUtils.LOWER_CHANNEL_LIMIT.put(UCentralConstants.BAND_5G, 36);
-	}
-
-	/** Map of band to the band-specific highest available channel*/
-	public static final Map<String, Integer> UPPER_CHANNEL_LIMIT =
-		new HashMap<>();
-	static {
-		UCentralUtils.UPPER_CHANNEL_LIMIT.put(UCentralConstants.BAND_2G, 11);
-		UCentralUtils.UPPER_CHANNEL_LIMIT.put(UCentralConstants.BAND_5G, 165);
-	}
-
-	/** List of available channels per band for use. */
-	public static final Map<String, List<Integer>> AVAILABLE_CHANNELS_BAND = Collections
-		.unmodifiableMap(buildBandToChannelsMap());
+	/** Map from band to ordered (increasing) list of available channels */
+	public static final Map<String, List<Integer>> AVAILABLE_CHANNELS_BAND =
+		Collections
+			.unmodifiableMap(buildBandToChannelsMap());
 
 	// This class should not be instantiated.
 	private UCentralUtils() {}
-	
+
+	/**
+	 * Builds map from band to ordered (increasing) list of available channels.
+	 */
 	private static Map<String, List<Integer>> buildBandToChannelsMap() {
 		Map<String, List<Integer>> bandToChannelsMap = new HashMap<>();
 		bandToChannelsMap.put(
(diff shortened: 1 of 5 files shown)


owrrm/src/test/java/com/facebook/openwifi/rrm/optimizers/TestUtils.java
```diff
@@ -187,17 +165,18 @@ public static JsonArray createDeviceStatus(List<String> bands) {
 	}
 
 	/**
-	 * Create an array with one radio info entry with the given tx power and
-	 * channel.
+	 * Create an array with one radio info entry with the given bandm, channel,
```
pohanhf, line 168: bandm -> band

pohanhf (approved):
Thanks Rocky for quickly adding the 6G support for channel assignment and TPC! The logics/data look good. It would be great if we could add some BAND_6G unit tests for those channel assignment algorithms to complete the coverage. Thanks.

owrrm/src/main/java/com/facebook/openwifi/rrm/optimizers/channel/ChannelOptimizer.java
```diff
@@ -647,7 +728,7 @@ protected void logPerfMetrics(
 	 * @param configManager the ConfigManager instance
 	 * @param channelMap the map of devices (by serial number) to radio to channel
 	 */
-	public void updateDeviceApConfig(
+	public static void updateDeviceApConfig(
```
elludraon, line 731: Please keep this consistent. If you really want to change this, update all callers, and do the same for `TPC`. Otherwise leave it as is (there's not really any benefit to making it static).

elludraon (approved):

owrrm/src/main/java/com/facebook/openwifi/rrm/optimizers/channel/ChannelOptimizer.java
RockyMandayam2: Its caller (same caller for TPC) cannot be made static, but I will update the TPC method as well.

pohanhf (approved):
Thanks for addressing the comments and adding the unit tests for 6G. :) It looks good to me.

lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/UCentralUtils.java
```diff
@@ -422,20 +482,61 @@ public static Map<String, String> getBssidsMap(
 	 * Determines if the given channel is in the given band.
 	 *
 	 * @param channel channel number
-	 * @param band    "2G" or "5G"
+	 * @param band band (e.g., "2G")
 	 * @return true if the given channel is in the given band; false otherwise
 	 */
 	public static boolean isChannelInBand(int channel, String band) {
-		return AVAILABLE_CHANNELS_BAND.get(band).contains(channel);
+		List<Integer> channels = AVAILABLE_CHANNELS_BAND.get(band);
```
paulvarkey, line 489: This assumes that channels is sorted -- is this tested elsewhere

lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/UCentralUtils.java
```diff
@@ -422,20 +482,61 @@ public static Map<String, String> getBssidsMap(
...
+		if (band.equals(UCentralConstants.BAND_2G)) {
+			if (channel == 14) {
+				// special case
+				return 2484;
+			} else {
+				return 2407 + 5 * channel;
+			}
+		} else if (band.equals(UCentralConstants.BAND_5G)) {
+			return 5000 + 5 * channel;
 		} else {
```
paulvarkey, line 512: because this is a public API, do you want to explicitly check for 6G for this case, and then do an error case with the final else?

lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/UCentralUtils.java
```diff
@@ -422,20 +482,61 @@ public static Map<String, String> getBssidsMap(
...
+	}
+
+	/**
+	 * Return the band that contains the given frequency (MHz).
+	 *
+	 * @param freqMHz center frequency (MHz)
+	 * @return the band that contains the given frequency, or null if no band
+	 *         contains it.
+	 */
+	public static String freqToBand(int freqMHz) {
```
paulvarkey, line 525: not important now -- but just compute and store this as class state, if this method is frequently called

owrrm/src/test/java/com/facebook/openwifi/rrm/optimizers/tpc/LocationBasedOptimalTPCTest.java
```diff
@@ -141,6 +144,10 @@ void testLocationBasedOptimalTPCSuccessCase() throws Exception {
 						DEFAULT_CHANNEL_5G,
 						DEFAULT_CHANNEL_WIDTH,
 						DEFAULT_TX_POWER,
+						dummyBssid,
```
paulvarkey, line 147: this State the DeviceCapability below can be created once and used twice, if you want to save some lines of code -- might make it more readable (your call)

paulvarkey (changes requested):

PR #108, opened 2022-10-20
Create separate Capabilities structure 

jwoos:
gson is pretty heavy - I was able to reduce 75% of memory usage by moving to a class we defined (basically a plain old data structure).

I used JOL to get footprint and used the same place to print the footprint:
- before: 86528
- after: 20024

Renamed `DeviceCapability` to `Capability` for the one in AP since it clashed with the one in GW. Also renamed the attribute to better reflect that it doesn't hold the whole capabilities object but just the phy portion


lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/models/ap/Capabilities.java
```diff
@@ -0,0 +1,47 @@
...
+	public Map<String, String[]> network;
+
+	public static class Switch {
+		public boolean enable;
+		public boolean reset;
+	}
+
+	@SerializedName("switch") public Map<String, Switch> switch_;
+
+	public static class Phy {
```
elludraon, line 28: You're missing a few fields:
- `country` (string)
- `dfs_region` (integer)
- `temperature` (integer)

Sources:
- https://github.com/blogic/ucentral-wifi/blob/aa9c3ee716929d75852a11640362a09bb32c5262/nl80211.c#L787
- https://github.com/blogic/ucentral-wifi/blob/aa9c3ee716929d75852a11640362a09bb32c5262/nl80211.h#L15

lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/models/ap/Capabilities.java
```diff
@@ -0,0 +1,47 @@
...
+		public String[] band;
+		public int ht_capa;
+		public int vht_capa;
+		public Integer[] he_phy_capa;
+		public Integer[] he_mac_capa;
+	}
+
+	public Map<String, Phy> wifi;
+	// The fields below were omitted
+	// macaddr;
```
elludraon, line 44: This might be `Map<String, String>` but unsure about the value type, leaving it commented is fine

lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/models/ap/Capabilities.java
```diff
@@ -0,0 +1,47 @@
...
+		public String[] htmode;
+		public String[] band;
+		public int ht_capa;
+		public int vht_capa;
+		public Integer[] he_phy_capa;
+		public Integer[] he_mac_capa;
+	}
+
+	public Map<String, Phy> wifi;
+	// The fields below were omitted
```
elludraon, line 43: put TODO here

elludraon (approved):
Looks mostly good. Accepting to unblock, please address comments.

.gitignore
```diff
@@ -14,3 +14,5 @@ bin/
 *.iml
 *.iws
 *.ipr
+
+temp/
```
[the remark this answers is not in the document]
jwoos: just for miscellaneous files to be placed into! for my use case specifically putting the settings file in there and random scripts

lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/models/ap/Capabilities.java
jwoos: oh wow good to know - i was looking at https://github.com/Telecominfraproject/wlan-ucentral-schema/blob/bfcb0e8c6e6c07195ba964f696c562aae3723fa3/tests/mocks/ubus/wifi%7Ephy.json

PR #110, opened 2022-10-21
Make separate uCentral schema structure

jwoos:
Create a separate class for UCentral schema - this will avoid gson and heavy memory usage.

Before: 8704
After: 2640


elludraon (changes requested):
Blocking due to T130685894 (sorry)

PR #111, opened 2022-10-26
Move running image to ibm-semeru and introduce runner script

jwoos:
- adoptopenjdk is being dissolved, so users should move to eclipse temurin or ibm semeru. Since the former is hotspot based and we want openj9, use ibm semeru JUST FOR RUNTIME. We will still use eclipse temurin for building (although i think we can technically use semeru for building as well - don't have as much context here)
  - a benefit of this is that there's aarch64 build so we can build/run native docker images on the ARM macbooks
- Add `runner.sh` to mirror running both in docker and locally. This will help speed up performance tuning

### DIFF
--- a/runner.sh
+++ b/runner.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env sh
+
+BIN=$1
+shift 1
+
+echo $@
+
+java \
+	-XX:+IdleTuningGcOnIdle -Xtune:virtualized \
+	-jar "$BIN" \
+	run "$@"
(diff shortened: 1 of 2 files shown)


runner.sh
```diff
@@ -0,0 +1,11 @@
...
+
+BIN=$1
+shift 1
+
+echo $@
+
+java \
+	-XX:+IdleTuningGcOnIdle -Xtune:virtualized \
+	-jar "$BIN" \
+	run "$@"
```
elludraon, line 11: can you move "run" back into the Dockerfile? (leave this generic, not specific to our RRM CLI)

elludraon (approved):
Thanks. Wow, no idea this happened to OpenJ9.

PR #112, opened 2022-10-27
JVM options for strings

jwoos:
Modify runner script to support different java binaries.

- Compact strings for both (default on openj9, but make it explicit)
- string deduplication for hotspot (seems to reduce a tiny amount of memory and g1 was already default gc so just make it explicity)

### DIFF
--- a/runner.sh
+++ b/runner.sh
@@ -7,7 +7,9 @@ JVM=$1
 BIN=$2
 shift 2
 
-COMMON_PARAMETERS=()
+COMMON_PARAMETERS=(
+-XX:+CompactStrings
+)
 
 if [ -n "$JVM_HOTSPOT" ]; then
 # for hotspot
--- a/runner.sh
+++ b/runner.sh
@@ -1,11 +1,29 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
 
-BIN=$1
-shift 1
+# script to help run the binary with specific jvm and jvm options
+# Defaults to openj9. Turn on hotspot by setting env JVM_HOTSPOT
 
-echo $@
+JVM=$1
+BIN=$2
+shift 2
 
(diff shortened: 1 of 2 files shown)


runner.sh
```diff
@@ -1,11 +1,33 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
```
elludraon, line 1: nit: you don't need bash arrays for this, use regular vars and backslash to escape newlines. I'm sure there is a use case for running in a smaller Docker container without bash.

runner.sh
```diff
@@ -1,11 +1,33 @@
...
 
-BIN=$1
-shift 1
+# script to help run the binary with specific jvm and jvm options
+# Defaults to openj9. Turn on hotspot by setting env JVM_HOTSPOT
```
elludraon, line 4: not really a fan of this - would prefer something like `JVM_IMPL` as input:
```shell
JVM_IMPL=${JVM_IMPL:-openj9}

if [ "$JVM_IMPL" = "openj9" ]; then
  # ...
else # "$JVM_IMPL" = "hotspot"
  # ...
fi
```

runner.sh
```diff
@@ -1,11 +1,33 @@
...
 
-echo $@
+JVM=$1
+BIN=$2
```
elludraon, line 7: this would be less confusing as `JAVA_BIN` and `JAR`, respectively (JVM is also a BIN and not used to seeing "java" bin referred to as "JVM")

elludraon (approved):

runner.sh
jwoos: ah that makes so much sense - i was confused as to why it didn't work! will change

elludraon (approved):

PR #114, opened 2022-11-01
Add rca configs

RockyMandayam2:
Add rca Configs class

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,18 +3,19 @@ WORKDIR /usr/src/java
 COPY . .
 RUN mvn clean package -pl owrrm -am -DappendVersionString="$(./scripts/get_build_version.sh)"
 
-FROM adoptopenjdk/openjdk11-openj9:latest
-RUN apt-get update && apt-get install -y gettext-base wget
-RUN wget https://raw.githubusercontent.com/Telecominfraproject/wlan-cloud-ucentral-deploy/main/docker-compose/certs/restapi-ca.pem \
-    -O /usr/local/share/ca-certificates/restapi-ca-selfsigned.pem
+FROM ibm-semeru-runtimes:open-11-jre
+RUN apt-get update && apt-get install -y gettext-base
+ADD https://raw.githubusercontent.com/Telecominfraproject/wlan-cloud-ucentral-deploy/main/docker-compose/certs/restapi-ca.pem \
+  /usr/local/share/ca-certificates/restapi-ca-selfsigned.pem
 RUN mkdir /owrrm-data
 WORKDIR /usr/src/java
 COPY docker-entrypoint.sh /
+COPY runner.sh /
 COPY --from=build /usr/src/java/owrrm/target/openwifi-rrm.jar /usr/local/bin/
 EXPOSE 16789 16790
 ENTRYPOINT ["/docker-entrypoint.sh"]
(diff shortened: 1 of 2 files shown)


jwoos (changes requested):
A lot of logs and configs that I don't think you meant to add!

jwoos (commented):
Looks good re: structure but since I'm not sure if the fields encompass everything that's needed I'll let someone else take a look as well.

jwoos (approved):

lib-rca/src/main/java/com/facebook/openwifi/librca/inputs/Configs.java
```diff
@@ -0,0 +1,174 @@
...
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+package com.facebook.openwifi.librca.inputs;
+
+/** Define root cause analysis configuration parameters */
+public final class Configs {
```
elludraon, line 12: can you rename to something more specific like "RCAParams" ?

lib-rca/src/main/java/com/facebook/openwifi/librca/inputs/Configs.java
```diff
@@ -0,0 +1,174 @@
...
+
+	/** Default look-back window in minutes */
+	public static final int DEFAULT_DETECTION_WINDOW = 360;
+
+	// Default KPI calculation parameters
+	/** Default minimum acceptable estimated throughput (Mbps) */
+	public static final double DEFAULT_MIN_ESTIMATED_THROUGHPUT = 10;
```
elludraon, line 19: Is there a reason you need to actually define all these consts? Do you plan to reference them later? Otherwise it's a lot of duplicate code/comments.

lib-rca/src/main/java/com/facebook/openwifi/librca/inputs/Configs.java
```diff
@@ -0,0 +1,174 @@
...
+	/** Default minimum acceptable AP bandwidth (MHz) for non-2G bands */
+	public static final int DEFAULT_MIN_AP_BANDWIDTH = 80;
+	/** Default minimum acceptable self airtime ratio (units are %) */
+	public static final double DEFAULT_MIN_SELF_AIRTIME_RATIO = 25.0;
+	/** Default maximum acceptable tx dropped ratio (units are %) */
+	public static final double DEFAULT_MAX_TX_DROPPED_RATIO = 0.1;
+
+
+	/** Look-back window in minutes */
+	public final int detectionWindow;
```
elludraon, line 64: all of these variable names should include the units (as a suffix), particularly time units

lib-rca/src/main/java/com/facebook/openwifi/librca/inputs/Configs.java
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	private static void validatePercentile(String varName, double value) {
+		if (value < 0 || value > 100) {
+			throw new IllegalArgumentException(
+				varName + " must be between 0 and 100 inclusive.");
+		}
+	}
+
+	public void validate() {
```
elludraon, line 150: - add comment
- also, you might want to consider returning a list of errors instead, as throwing will only let you show the first error

elludraon (changes requested):

PR #115, opened 2022-11-01
Add librca util function to calculate percentiles

RockyMandayam2:
Add librca util function to calculate percentiles.


lib-rca/src/test/java/com/facebook/openwifi/librca/UtilsTest.java
```diff
@@ -0,0 +1,111 @@
...
+		exp = values.get(500);
+		assertEquals(actual, exp);
+		actual = Utils.getValueAtPercentile(values, 99.9);
+		exp = values.get(999);
+		assertEquals(actual, exp);
+		// test ties
+		actual = Utils.getValueAtPercentile(values, 99.85);
+		int exp1 = values.get(998);
+		int exp2 = values.get(999);
+		assertTrue(actual == exp1 || actual == exp2);
```
jwoos, line 109: Does this not have a consistent behavior?

jwoos (changes requested):
please remove all the extra files!

paulvarkey (changes requested):
Was deletion of RootCauseAnalyzerTest.java intentional?

The test seems it ought to be deterministic -- so I second jwoos's comment.

lib-rca/src/test/java/com/facebook/openwifi/librca/UtilsTest.java
```diff
@@ -0,0 +1,111 @@
...
+
+		// test more values
+		values.clear();
+		for (int i = 0; i < 1000; i++) {
+			values.add(i);
+		}
+		actual = Utils.getValueAtPercentile(values, 50);
+		exp = values.get(500);
+		assertEquals(actual, exp);
+		actual = Utils.getValueAtPercentile(values, 99.9);
```
zhiqiand, line 102: Can you test a 10th/ 90th percentile? That's the only two use cases in RCA.

RockyMandayam2:
Closed this PR. Will open a new one using an apache library to do the calculations

PR #118, opened 2022-11-01
Refactor AggregatedState and add timestamp

zhiqiand:

This PR refactor `AggregatedState` to meet the input data standard of RCA. The input data structure of RCA will be defined later in a separate PR.

Quick review:
```
public class AggregatedState {   
   public static class RadioConfig {
       public int channel;
       public int channelWidth;
       public int txPower;
       public String phy;
   }
   public class AssociationInfo {
       public long activeMsRadio;
       public long busyMsRadio;
       public long noiseRadio;
       public long receiveMsRadio;
       public long transmitMsRadio;
 
       public long connected;
       public long inactive;
       public int rssi;
       public long rxBytes;
       public long rxPackets;
       public Rate rxRate;
       public long txBytes;
       public long txDuration;
       public long txFailed;
       public long txPackets;
       public Rate txRate;
       public long txRetries;
       public int ackSignal;
       public int ackSignalAvg;
       public long txPacketsCounters;
       public long txErrorsCounters;
       public long txDroppedCounters;
       public long timeStamp;
   }
 
   public String bssid;
   public String station;
   public RadioConfig radioConfig;
   public List<AssociationInfo> associationInfoList;
}
```

### DIFF
--- a/lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/AggregatedState.java
+++ b/lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/AggregatedState.java
@@ -22,34 +22,6 @@ import com.google.gson.JsonObject;
  */
 public class AggregatedState {
 
-	/** Rate information with aggregated fields. */
-	public static class Rate {
-		/**
-		 * Aggregated fields bitRate
-		 */
-		public long bitRate;
-
-		/**
-		 * Aggregated fields chWidth
-		 */
-		public int chWidth;
-
-		/**
-		 * Aggregated fields mcs
-		 */
-		public int mcs;
-
-		/** Constructor with no args */
-		private Rate() {}
-
-		/** Constructor with args */
-		private Rate(long bitRate, int chWidth, int mcs) {
-			this.bitRate = bitRate;
-			this.chWidth = chWidth;
-			this.mcs = mcs;
-		}
-	}
-
 	/**
 	 * Radio information with channel, channel_width and tx_power.
 	 */
@@ -99,11 +71,39 @@ public class AggregatedState {
 		}
 	}
 
-	/** 
-	 * Data model to keep raw data from {@link State.Interface.SSID.Association}, 
+	/**
+	 * Data model to keep raw data from {@link State.Interface.SSID.Association},
 	 * {@link State.Radio} and {@link State.Interface.Counters}.
 	 */
 	public static class AssociationInfo {
+		/** Rate information with aggregated fields. */
+		public static class Rate {
+			/**
+			 * Aggregated fields bitRate
+			 */
+			public long bitRate;
+
+			/**
+			 * Aggregated fields chWidth
+			 */
+			public int chWidth;
+
+			/**
+			 * Aggregated fields mcs
+			 */
+			public int mcs;
+
+			/** Constructor with no args */
+			private Rate() {}
+
+			/** Constructor with args */
+			private Rate(long bitRate, int chWidth, int mcs) {
+				this.bitRate = bitRate;
+				this.chWidth = chWidth;
+				this.mcs = mcs;
+			}
+		}
+
 		public long connected;
 		public long inactive;
 		public int rssi;
@@ -126,7 +126,9 @@ public class AggregatedState {
 		public long noiseRadio;
 		public long receiveMsRadio;
 		public long transmitMsRadio;
-		public long timeStamp;
+
+		/** unix time in ms */
+		public long timestamp;
 
 		/** Default Constructor. */
 		public AssociationInfo() {}
@@ -141,7 +143,7 @@ public class AggregatedState {
 			Association association,
 			Counters counters,
 			JsonObject radio,
-			long timeStamp
+			long timestamp
 		) {
 			// Association info
 			connected = association.connected;
@@ -187,16 +189,19 @@ public class AggregatedState {
 			receiveMsRadio = radio.get("receive_ms").getAsLong();
 			noiseRadio = radio.get("noise").getAsLong();
 
-			this.timeStamp = timeStamp;
+			this.timestamp = timestamp;
 		}
 	}
 
+	/** Aggregate AssociationInfo over bssid, station and RadioConfig */
 	public String bssid;
 	public String station;
 	public RadioConfig radioConfig;
+
+	/** Store a list of AssociationInfo of the same link and radio configuration. */
 	public List<AssociationInfo> associationInfoList;
 
-	/** Constructor with no args */
+	/** Constructor with no args. For test purpose. */
 	public AggregatedState() {
 		this.associationInfoList = new ArrayList<>();
 		this.radioConfig = new RadioConfig();
@@ -207,13 +212,13 @@ public class AggregatedState {
 		Association association,
 		Counters counters,
 		JsonObject radio,
-		long timeStamp
+		long timestamp
 	) {
 		this.bssid = association.bssid;
 		this.station = association.station;
 		this.associationInfoList = new ArrayList<>();
 		associationInfoList
-			.add(new AssociationInfo(association, counters, radio, timeStamp));
+			.add(new AssociationInfo(association, counters, radio, timestamp));
 		this.radioConfig = new RadioConfig(radio);
 	}
 
--- a/lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/AggregatedState.java
+++ b/lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/AggregatedState.java
@@ -193,12 +193,12 @@ public class AggregatedState {
 		}
 	}
 
-	/** Aggregate AssociationInfo over bssid, station and RadioConfig */
+	//Aggregate AssociationInfo over bssid, station and RadioConfig.
 	public String bssid;
 	public String station;
 	public RadioConfig radioConfig;
 
-	/** Store a list of AssociationInfo of the same link and radio configuration. */
+	//Store a list of AssociationInfo of the same link and radio configuration. */
 	public List<AssociationInfo> associationInfoList;
 
 	/** Constructor with no args. For test purpose. */
--- a/lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/AggregatedState.java
+++ b/lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/AggregatedState.java
@@ -10,11 +10,11 @@ package com.facebook.openwifi.cloudsdk;
 
 import java.util.ArrayList;
 import java.util.List;
-import java.util.Map;
 import java.util.Objects;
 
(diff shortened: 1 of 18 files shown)


lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/models/ap/State.java
```diff
@@ -132,4 +132,5 @@ public static class Radio {
 	@SerializedName("link-state") public JsonObject linkState;
 	public JsonObject gps;
 	public JsonObject poe;
+	public long timeStamp;
```
elludraon, line 135: this should NOT be added here if it isn't present in the model coming from OWF - define a subclass like we did for wifiscans (`WifiScanEntry` vs. `WifiScanEntryResult`)

lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/AggregatedState.java
```diff
@@ -23,75 +23,59 @@
...
 		public int txPower;
+		public String phy;
 
-		private Radio() {}
+		/** Default constructor with no args */
+		private RadioConfig() {}
 
-		public Radio(int channel, int channelWidth, int txPower) {
+		/** Constructor with args */
+		public RadioConfig(JsonObject radio) {
```
elludraon, line 66: Why are you passing in as `JsonObject`? We are trying to remove that where possible - you can define a new class for it, or pass individual fields if you have to

lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/AggregatedState.java
```diff
@@ -109,64 +93,128 @@ public boolean equals(Object obj) {
...
+		public int ackSignalAvg;
+		public long txPacketsCounters;
+		public long txErrorsCounters;
+		public long txDroppedCounters;
+		public long activeMsRadio;
+		public long busyMsRadio;
+		public long noiseRadio;
+		public long receiveMsRadio;
+		public long transmitMsRadio;
+		public long timeStamp;
```
elludraon, line 129: nit: `timestamp` (no uppercase, it is one word)

elludraon (changes requested):

lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/AggregatedState.java
zhiqiand: I pass it in as `JsonObect` because [radio under `SSID` is of `JsonObject` type](https://github.com/Telecominfraproject/wlan-cloud-rrm/blob/88d5766bef61d4f3e190130312d074831a6f1e8f/lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/models/ap/State.java#L68). Although there is a `Radio` class under `State`. it's a Radio array which contains all the Radio observed in the State

elludraon: Ah ok, we can address this later then when also addressing it in `State`.

lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/AggregatedState.java
```diff
@@ -109,64 +93,128 @@ public boolean equals(Object obj) {
...
-	public AggregatedRate txRate;
-	public long txRetries;
-	public int ackSignal;
-	public int ackSignalAvg;
-	public Radio radio;
+	public RadioConfig radioConfig;
+	public List<AssociationInfo> associationInfoList;
 
 	/** Constructor with no args */
 	public AggregatedState() {
```
RockyMandayam2, line 200: Is this only used for testing? If so, please mention that in the javadoc.

RockyMandayam2 (approved):

lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/AggregatedState.java
```diff
@@ -109,64 +65,161 @@ public boolean equals(Object obj) {
...
+			busyMsRadio = radio.get("busy_ms").getAsLong();
+			transmitMsRadio = radio.get("transmit_ms").getAsLong();
+			receiveMsRadio = radio.get("receive_ms").getAsLong();
+			noiseRadio = radio.get("noise").getAsLong();
+
+			this.timestamp = timestamp;
+		}
+	}
+
+	/** Aggregate AssociationInfo over bssid, station and RadioConfig */
```
elludraon, line 196: don't put javadoc comments when they aren't tied to a specific item - use a regular comment

lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/models/ap/State.java
```diff
@@ -132,4 +302,24 @@ public static class Radio {
 	@SerializedName("link-state") public JsonObject linkState;
 	public JsonObject gps;
 	public JsonObject poe;
+
+	/** Default Constructor with no args */
+	public State() {}
+
+	/** Copy Constructor */
+	public State(State state) {
```
elludraon, line 310: I don't understand why you had to make all these changes here -- this makes editing and reading this file very difficult. Isn't this not even being used? If you absolutely need it, why not just deep copy via `gson` (does it need to be performant?)

elludraon (changes requested):

lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/models/ap/State.java
zhiqiand: Will do.

elludraon: I am still confused, where is this used? There is already a `Utils.deepCopy()` method too.

elludraon (commented):
Something went wrong in your merge, this diff is showing other commits

lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/models/ap/State.java
zhiqiand: Ah, I'm thinking of copying information from `State` to `StateInfo` when serializing a new `StateInfo` instance in `Moduler` .

lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/StateInfo.java
```diff
@@ -0,0 +1,22 @@
...
+
+public class StateInfo extends State {
+	/**
+	 * Unix time in milliseconds (ms). This is added it because {@link State} is an unknown
+	 * time reference.
+	 */
+	public long timestamp;
+
+	/** Default Constructor. */
+	public StateInfo() {}
```
elludraon, line 21: remove, this is implicit

lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/models/ap/State.java
```diff
@@ -132,4 +132,7 @@ public static class Radio {
 	@SerializedName("link-state") public JsonObject linkState;
 	public JsonObject gps;
 	public JsonObject poe;
+
+	/** Default Constructor with no args */
+	public State() {}
```
elludraon, line 137: remove, this is implicit

lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/StateInfo.java
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+package com.facebook.openwifi.cloudsdk;
+
+import com.facebook.openwifi.cloudsdk.models.ap.State;
+
+public class StateInfo extends State {
+	/**
+	 * Unix time in milliseconds (ms). This is added it because {@link State} is an unknown
+	 * time reference.
```
elludraon, line 16: nit: be more specific and write "State.unit.localtime" (without the `@link` annotation is fine)

owrrm/src/main/java/com/facebook/openwifi/rrm/modules/Modeler.java
```diff
@@ -277,7 +277,8 @@ private void fetchInitialData() {
 			JsonObject state = records.data.get(0).data;
 			if (state != null) {
 				try {
-					State stateModel = gson.fromJson(state, State.class);
+					StateInfo stateModel =
+						gson.fromJson(state, StateInfo.class);
```
elludraon, line 281: add a timestamp here too?

lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/UCentralUtils.java
```diff
@@ -387,11 +387,11 @@ public static Map<String, Map<String, List<Integer>>> getDeviceAvailableChannels
...
 	 */
 	public static Map<String, String> getBssidsMap(
-		Map<String, State> latestState
+		Map<String, StateInfo> latestState
 	) {
 		Map<String, String> bssidMap = new HashMap<>();
-		for (Map.Entry<String, State> e : latestState.entrySet()) {
-			State state = e.getValue();
+		for (Map.Entry<String, StateInfo> e : latestState.entrySet()) {
+			StateInfo state = e.getValue();
```
elludraon, line 394: You don't need to change this line if you don't actually need the subclass - keep the declaration as `State`. Same comment in the other files (`ModelerUtils` etc.) -- you don't need to change as many lines as you have.

Actually, you can make the least amount of changes if you modify the function parameters to use generics, like this:
```
Map<String, StateInfo>
-->
Map<String, ? extends State>  // only when you don't need StateInfo specific fields
```

owrrm/src/main/java/com/facebook/openwifi/rrm/modules/Modeler.java
zhiqiand: Yep, but I cannot figure out where the timestamp is.. Do you have any thoughts?

elludraon: Should be in `records.get(i).recorded`, but verify for yourself

lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/UCentralUtils.java
zhiqiand: In `ModelerUtils`, `StateInfo` specific fields are needed. I change it in other files. Thanks for pointing out!

elludraon (approved):
Looks good. Thanks for making the changes!

PR #122, opened 2022-11-03
Prepare for client steering

RockyMandayam2:
Prepare for client steering: add simple logic to the cloud and make it callable (one-off or scheduled, still no trigger detection), but the actual steering of clients is not implemented yet.

### DIFF
--- a/owrrm/src/main/java/com/facebook/openwifi/rrm/RRMAlgorithm.java
+++ b/owrrm/src/main/java/com/facebook/openwifi/rrm/RRMAlgorithm.java
@@ -20,6 +20,8 @@ import com.facebook.openwifi.rrm.optimizers.channel.ChannelOptimizer;
 import com.facebook.openwifi.rrm.optimizers.channel.LeastUsedChannelOptimizer;
 import com.facebook.openwifi.rrm.optimizers.channel.RandomChannelInitializer;
 import com.facebook.openwifi.rrm.optimizers.channel.UnmanagedApAwareChannelOptimizer;
+import com.facebook.openwifi.rrm.optimizers.clientsteering.ClientSteeringOptimizer;
+import com.facebook.openwifi.rrm.optimizers.clientsteering.SingleAPBandSteering;
 import com.facebook.openwifi.rrm.optimizers.tpc.LocationBasedOptimalTPC;
 import com.facebook.openwifi.rrm.optimizers.tpc.MeasurementBasedApApTPC;
 import com.facebook.openwifi.rrm.optimizers.tpc.MeasurementBasedApClientTPC;
@@ -42,6 +44,10 @@ public class RRMAlgorithm {
 		OptimizeTxPower (
 			"Optimize tx power configuration",
 			"Run transmit power control algorithm"
+		),
+		ClientSteering (
+			"Steer clients onto the optimal AP and band",
+			"Run client steering algorithm"
 		);
 
 		/** The long name. */
@@ -72,6 +78,13 @@ public class RRMAlgorithm {
 		 * @see TPC#computeTxPowerMap()
 		 */
 		public Map<String, Map<String, Integer>> txPowerMap;
+
+		/**
+		 * Computed actions for each AP-client pair.
+		 *
+		 * @see ClientSteeringOptimizer#computeApClientActionMap()
+		 */
+		public Map<String, Map<String, String>> apClientActionMap;
 	}
 
 	/** The algorithm name (should be AlgorithmType enum string). */
@@ -282,6 +295,39 @@ public class RRMAlgorithm {
 					configManager,
 					result.txPowerMap
 				);
+				if (updateImmediately) {
+					configManager.queueZoneAndWakeUp(zone);
+				}
+			}
+		} else if (
+			name.equals(RRMAlgorithm.AlgorithmType.ClientSteering.name())
+		) {
+			logger.info(
+				"Zone '{}': Running client steering optimizer (mode='{}')",
+				zone,
+				mode
+			);
+			ClientSteeringOptimizer optimizer;
+			switch (mode) {
+			default:
+				if (!allowDefaultMode || !mode.isEmpty()) {
+					result.error = modeErrorStr;
+					return result;
+				}
+				logger.info("Using default algorithm mode...");
+				// fall through
+			case SingleAPBandSteering.ALGORITHM_ID:
+				optimizer = SingleAPBandSteering.makeWithArgs(
+					modeler.getDataModelCopy(),
+					zone,
+					deviceDataManager,
+					args
+				);
+				break;
+			}
+			result.apClientActionMap = optimizer.computeApClientActionMap();
+			if (!dryRun) {
+				optimizer.steer(result.apClientActionMap);
 				if (updateImmediately) {
 					configManager.queueZoneAndWakeUp(zone);
 				}
--- a/owrrm/src/main/java/com/facebook/openwifi/rrm/RRMAlgorithm.java
+++ b/owrrm/src/main/java/com/facebook/openwifi/rrm/RRMAlgorithm.java
@@ -86,7 +86,7 @@ public class RRMAlgorithm {
 		/**
 		 * Computed actions for each AP-client pair.
 		 *
-		 * @see ClientSteeringOptimizer#computeApClientActionMap()
+		 * @see ClientSteeringOptimizer#computeApClientActionMap(boolean)
 		 */
 		public Map<String, Map<String, String>> apClientActionMap;
 	}
@@ -330,7 +330,8 @@ public class RRMAlgorithm {
 				);
 				break;
 			}
-			result.apClientActionMap = optimizer.computeApClientActionMap();
+			result.apClientActionMap =
+				optimizer.computeApClientActionMap(dryRun);
 			if (!dryRun) {
 				optimizer.steer(result.apClientActionMap);
 				if (updateImmediately) {
--- a/owrrm/src/main/java/com/facebook/openwifi/rrm/RRMAlgorithm.java
+++ b/owrrm/src/main/java/com/facebook/openwifi/rrm/RRMAlgorithm.java
@@ -36,6 +36,11 @@ public class RRMAlgorithm {
 	private static final Logger logger =
 		LoggerFactory.getLogger(RRMAlgorithm.class);
 
+	/**
+	 * Client steering state. A single instance must be passed into all client
+	 * steering algorithms, as the state must persist across runs of the
+	 * optimizers.
+	 */
 	private static final ClientSteeringState clientSteeringState =
 		new ClientSteeringState();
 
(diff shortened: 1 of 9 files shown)


owrrm/src/main/java/com/facebook/openwifi/rrm/optimizers/clientsteering/ClientSteeringState.java
```diff
@@ -0,0 +1,111 @@
...
+
+package com.facebook.openwifi.rrm.optimizers.clientsteering;
+
+import java.util.Map;
+import java.util.TreeMap;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+/** Singleton class to manage global client steering state */
+public final class ClientSteeringState {
```
halmehmood, line 17: The state is supposed to be per client, per AP

elludraon: Yeah, also confused about `apRadioLastAttempt` - it should be AP -> STA, not AP -> radio?

owrrm/src/main/java/com/facebook/openwifi/rrm/optimizers/clientsteering/ClientSteeringState.java
```diff
@@ -0,0 +1,111 @@
...
+
+	/**
+	 * Register the time of the latest client steering attempt by the given AP
+	 * and radio.
+	 *
+	 * @param apSerialNumber non-null AP serial number
+	 * @param bssid non-null radio bssid
+	 * @param currentTimeNs current JVM monotonic time (ns)
+	 */
+	final void registerClientSteeringAttempt(
```
elludraon, line 52: - Why are all of these `final`?
- Missing `public`?
- You need to think about how multiple concurrent `ClientSteeringOptimizer` instances might be calling these methods - hopefully not on the same AP simultaneously, but you may either need to handle that case or explicitly disallow it.

owrrm/src/main/java/com/facebook/openwifi/rrm/optimizers/clientsteering/ClientSteeringState.java
```diff
@@ -0,0 +1,111 @@
...
+		String apSerialNumber,
+		String bssid,
+		long currentTimeNs
+	) {
+		Map<String, Long> radioLastAttempt = apRadioLastAttempt
+			.computeIfAbsent(apSerialNumber, k -> new TreeMap<>());
```
elludraon, line 58: again, why `TreeMap`? that should not be your default unless you need the keys in sorted order - it's `O(log n)` while `HashMap` is `O(1)`

owrrm/src/main/java/com/facebook/openwifi/rrm/optimizers/clientsteering/SingleAPBandSteering.java
```diff
@@ -0,0 +1,286 @@
...
+	/** RSSI below which 2G clients are deauthenticated */
+	private final short minRssi2G;
+	/** RSSI above which 2G clients are asked to move to 5G or 6G */
+	private final short maxRssi2G;
+	/** RSSI below which 5G and 6G clients are asked to move to 2G */
+	private final short minRssiNon2G;
+	/** Backoff time (ms) for all APs and radios */
+	private final int backoffTime;
+
+	public static SingleAPBandSteering makeWithArgs(
```
elludraon, line 70: add comment

owrrm/src/main/java/com/facebook/openwifi/rrm/optimizers/clientsteering/ClientSteeringState.java
RockyMandayam2: They don't have to be final, wasn't sure about that. I didn't want it public as I figured nothing outside this package would need to be considered with this (I was going to remove `public` from the other methods too, i forgot).

Regarding concurrency, I'm not sure the best way to resolve this. Does it make sense to just make `ClientSteeringOptimizer#computeApClientActionMap` `synchronized` and make sure (and leave a comment) not to access `ClientSteeringState#registerClientSteeringAttempt` anywhere except in implementations of that method?

elludraon: IMO keep `public` and drop the `final`. Don't overcomplicate this unless there is a need.

For the methods, I am more concerned about interleaving "getLatest", "checkBackoff", "register" methods from the caller. Think about "transactional" operations, which might mean changing the methods or doing locking on keys (AP, client, etc.).

owrrm/src/test/java/com/facebook/openwifi/rrm/optimizers/clientsteering/SingleAPBandSteeringTest.java
```diff
@@ -185,6 +185,7 @@ void testComputeApClientActionMap() {
 		);
 		Map<String, Map<String, String>> apClientActionMap =
 			optimizer.computeApClientActionMap();
+		System.out.println(exp);
```
elludraon, line 188: delete (I usually add a `TODO` comment on debug logs so I can find them easily later)

owrrm/src/main/java/com/facebook/openwifi/rrm/optimizers/clientsteering/ClientSteeringOptimizer.java
```diff
@@ -0,0 +1,81 @@
...
+		this.model.latestDeviceStatusRadios.keySet()
+			.removeIf(serialNumber -> !deviceConfigs.containsKey(serialNumber));
+		this.model.latestDeviceCapabilitiesPhy.keySet()
+			.removeIf(serialNumber -> !deviceConfigs.containsKey(serialNumber));
+	}
+
+	/** Get global client steering state */
+	protected final ClientSteeringState getClientSteeringState() {
+		return ClientSteeringState.getInstance();
+	}
```
elludraon, line 62: IMO you should be passing `ClientSteeringState` via the constructor or in another method. Then you also don't need that test-only `reset()` method (make `ClientSteeringState` constructor public and create a new instance).

owrrm/src/main/java/com/facebook/openwifi/rrm/optimizers/clientsteering/ClientSteeringState.java
```diff
@@ -0,0 +1,89 @@
...
+	 * values are never null.
+	 */
+	private ConcurrentMap<String, Map<String, Long>> apClientLastAttempt =
+		new ConcurrentHashMap<>();
+
+	/** Reset the state - ONLY FOR TESTING */
+	public synchronized void reset() {
+		// TODO better way then to call this before every test??
+		apClientLastAttempt.clear();
+	}
```
elludraon, line 43: see other comment

owrrm/src/main/java/com/facebook/openwifi/rrm/optimizers/clientsteering/ClientSteeringState.java
```diff
@@ -0,0 +1,89 @@
...
+	 * Note that if there was a previous attempt for the given AP and station,
+	 * the current time must not be before the last attempt.
+	 *
+	 * @param apSerialNumber AP serial number
+	 * @param station client MAC
+	 * @param currentTimeNs JVM monotonic time in ns
+	 * @param backoffTimeNs non-negative backoff time (ns)
+	 * @return true if client steering attempt was registered; false otherwise
+	 */
+	public synchronized boolean registerIfBackoffExpired(
```
elludraon, line 64: Why synchronize on this outer class? That is defeating the purpose of your `ConcurrentHashMap` (as it will never be referenced concurrently anyway). As a basic example, you could instead synchronize on `clientLastAttempt` below, i.e.
```
		// get last attempt
		Map<String, Long> clientLastAttempt = apClientLastAttempt
			.computeIfAbsent(apSerialNumber, k -> new HashMap<>());
		synchronized (clientLastAttempt) {
			Long lastAttempt = /* ... */
		}
```
which at least permits the "normal case" of parallel operations for *different APs*.

owrrm/src/main/java/com/facebook/openwifi/rrm/optimizers/clientsteering/SingleAPBandSteering.java
```diff
@@ -0,0 +1,272 @@
...
+					);
+					if (band == null) {
+						continue;
+					}
+					for (
+						State.Interface.SSID.Association assoc : ssid.associations
+					) {
+						// decide whether to do any band steering
+						// TODO check which bands AP & client can use (see 11k)
+						if (UCentralConstants.BAND_2G.equals(band)) {
```
elludraon, line 180: IMO this loop body is way too far nested and you should define a new method for this. (The code is line breaking too much to be readable)

owrrm/src/main/java/com/facebook/openwifi/rrm/optimizers/clientsteering/ClientSteeringState.java
```diff
@@ -0,0 +1,89 @@
...
+		new ConcurrentHashMap<>();
+
+	/** Reset the state - ONLY FOR TESTING */
+	public synchronized void reset() {
+		// TODO better way then to call this before every test??
+		apClientLastAttempt.clear();
+	}
+
+	/**
+	 * Register a client steering attempt for the given AP and station at the
```
elludraon, line 46: This is a little weird to me because you are registering the attempt in a different section of code from actually executing the attempt. That is fine, but please make it more clear that this is only checking the **registration time** not the actual execution time.

Also, looks like this doesn't work in the "dry run" situation. Please pass that flag down.

elludraon (changes requested):

owrrm/src/main/java/com/facebook/openwifi/rrm/RRMAlgorithm.java
```diff
@@ -35,6 +36,9 @@ public class RRMAlgorithm {
 	private static final Logger logger =
 		LoggerFactory.getLogger(RRMAlgorithm.class);
 
+	private static final ClientSteeringState clientSteeringState =
+		new ClientSteeringState();
```
elludraon, line 40: This is fine - can you add a comment that this is important and a only single instance needs to be passed to all algorithms? (This would have been obvious as a singleton, but otherwise you should add a note.)

owrrm/src/main/java/com/facebook/openwifi/rrm/optimizers/clientsteering/ClientSteeringState.java
```diff
@@ -13,20 +13,11 @@
...
-/** Singleton class to manage global client steering state */
-public final class ClientSteeringState {
-
-	/** Singleton instance */
-	private static final ClientSteeringState singleton =
-		new ClientSteeringState();
+/** Class to manage global client steering state */
+public class ClientSteeringState {
 
 	/** Private default constructor (this is a singleton class) */
```
elludraon, line 19: delete/update comment

elludraon (approved):

PR #123, opened 2022-11-03
Add rca params

RockyMandayam2:
Add rca params

### DIFF
--- a/lib-rca/src/main/java/com/facebook/openwifi/librca/inputs/RCAParams.java
+++ b/lib-rca/src/main/java/com/facebook/openwifi/librca/inputs/RCAParams.java
@@ -1,0 +1,128 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+package com.facebook.openwifi.librca.inputs;
+
+/** Define root cause analysis configuration parameters */
+public final class RCAParams {
+
+	/** Look-back window in minutes */
+	public final int detectionWindowMin;
+
+	// KPI calculation parameters
+	/** Minimum acceptable estimated throughput (Mbps) */
+	public final double minEstimatedThroughputMbps;
+	/** Percentile (units are %) of estimated throughputs to use as the KPI */
+	public final double throughputAggregationPercentile;
+	/** Maximum acceptable latency (ms) */
+	public final int maxLatencyThresholdMs;
+	/** Maximum acceptable jitter (ms) */
+	public final int maxJitterThresholdMs;
+	/**
+	 * Maximum acceptable disconnection rate (disconnetions per hour). Note that
+	 * this signifies a rate and the units happen to be per hour - this does not
+	 * signify that every contiguous one-hour period be checked.
+	 */
+	public final int maxDisconnectionRatePerHour;
+
+	// High Level metrics parameters
+	/** Minimum acceptable tx rate (Mbps) */
+	public final double minTxRateMbps;
+	/** Maximum acceptable Packet Error Rate (PER) (units are %) */
+	public final double maxPERPercent;
+	/** Minimum acceptable idle airtime (units are %) */
+	public final double minIdleAirtimePercent;
+	/** Maximum acceptable number of clients for one radio */
+	public final int maxNumClients;
+
+	// Low Level metrics parameters
+	/** Minimum acceptable RSSI (dBm) */
+	public final int minRssidBm;
+	/** Maximum acceptable noise (dBm) */
+	public final int maxNoisedBm;
+	/** Maximum acceptable intf airtime (units are %) */
+	public final double maxIntfAirtimePercent;
+	/** Maximum acceptable number of neighbors */
+	public final int maxNumNeighbors;
+	/** Minimum acceptable client bandwidth (MHz) for non-2G bands / */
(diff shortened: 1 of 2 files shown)


jwoos (approved):

lib-rca/src/main/java/com/facebook/openwifi/librca/inputs/RCAParams.java
```diff
@@ -0,0 +1,208 @@
...
+		this.minTxRateMbps = 50;
+		this.maxPERPercent = 10.0;
+		this.minIdleAirtimePercent = 10.0;
+		this.maxNumClients = 10;
+
+		this.minRssidBm = -70;
+		this.maxNoisedBm = -95;
+		this.maxIntfAirtimePercent = 75.0;
+		this.maxNumNeighbors = 10;
+		this.minClientBandwidthMHz = 80;
```
zhiqiand, line 83: There is also bandwidth rules for 2.4GHz, maybe it's better to add it as well.

lib-rca/src/main/java/com/facebook/openwifi/librca/inputs/RCAParams.java
```diff
@@ -0,0 +1,208 @@
...
+	/** Maximum acceptable jitter (ms) */
+	public final int maxJitterThresholdMs;
+	/**
+	 * Maximum acceptable disconnection rate (disconnetions per hour). Note that
+	 * this signifies a rate and the units happen to be per hour - this does not
+	 * signify that every contiguous one-hour period be checked.
+	 */
+	public final int maxDisconnectionRatePerHour;
+
+	// High Level metrics parameters
```
zhiqiand, line 36: `// High Level metrics parameters` -> `// High Level metrics threshold` 
Or something like that? They are parameters to compare with

lib-rca/src/main/java/com/facebook/openwifi/librca/inputs/RCAParams.java
```diff
@@ -0,0 +1,208 @@
...
+package com.facebook.openwifi.librca.inputs;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/** Define root cause analysis configuration parameters */
+public final class RCAParams {
+
+	/** Look-back window in minutes */
+	public final int detectionWindowMin;
```
zhiqiand, line 18: Why not use milliseconds? So that we don't need to transfer it to ms during calculation.
Just put `(6hr)` in your java doc, then it's clear enough to help developers understand how long the time window is.

lib-rca/src/main/java/com/facebook/openwifi/librca/inputs/RCAParams.java
```diff
@@ -0,0 +1,208 @@
...
+		String varName,
+		double value,
+		List<String> errors
+	) {
+		if (value < 0 || value > 100) {
+			errors.add(varName + " must be between 0 and 100 inclusive.");
+		}
+	}
+
+	public List<String> validate() {
```
elludraon, line 132: add javadoc

lib-rca/src/main/java/com/facebook/openwifi/librca/inputs/RCAParams.java
```diff
@@ -0,0 +1,208 @@
...
+	public final int minClientBandwidthMHz;
+	/** Minimum acceptable Access Point (AP) bandwidth (MHz) for non-2G bands */
+	public final int minApBandwidthMHz;
+	/** Minimum acceptable self airtime ratio (units are %) */
+	public final double minSelfAirtimeRatioPercent;
+	/** Maximum acceptable tx dropped ratio (units are %) */
+	public final double maxTxDroppedRatioPercent;
+
+	/** Default constructor */
+	public RCAParams() {
```
elludraon, line 65: Don't you need another constructor? Or remove "final" keyword from everything here?

lib-rca/src/main/java/com/facebook/openwifi/librca/inputs/RCAParams.java
RockyMandayam2: I can do ms, I'll change it

lib-rca/src/main/java/com/facebook/openwifi/librca/inputs/RCAParams.java
RockyMandayam2: Sure, and I can do the same with the low level ones also

lib-rca/src/main/java/com/facebook/openwifi/librca/inputs/RCAParams.java
RockyMandayam2: I figured the input would eventually come in as json, so when we parse that, we wouldn't need a separate constructor. (For now, we'd just use the default values for everything.) I also wanted to keep final just so they can be public without adding getters. Thoughts?

elludraon: Can you add that as a comment on the class?

lib-rca/src/main/java/com/facebook/openwifi/librca/inputs/RCAParams.java
RockyMandayam2: Haleema said channel bonding is not common/recommended in 2G, so we would expect the smallest (normal) channel width always in 2G, so no need to add a parameter for min bandwidth in 2G.

elludraon (approved):

zhiqiand (approved):

PR #124, opened 2022-11-03
Add CDS for hotspot and ability to add extra JVM flags in runner

jwoos:
1. Enabled App CDS (Class Data Sharing) with the hopes that I'd see some performance improvement (no such luck - at least in memory).
2. Added ability to pass in extra JVM flags in runner.sh - this is useful


RockyMandayam2 (approved):
Thanks!

PR #125, opened 2022-11-03
Data Models for RCA

zhiqiand:
This PR creates data models for RCA, including `InputStats`, `LinkStats` and `ClientStats`. 
Reference: https://docs.google.com/document/d/1smWAN-Dq4gAYHGFX5AEEcgSn9ilIPQviCWFqOEAHmDs/edit?usp=sharing


lib-rca/src/main/java/com/facebook/openwifi/librca/models/ClientStats.java
```diff
@@ -0,0 +1,12 @@
+package com.facebook.openwifi.librca.models;
```
RockyMandayam2, line 1: Are all three of these classes aggregated info/statistics? If so, I think the package name should be `aggregatedstats`?

lib-rca/src/main/java/com/facebook/openwifi/librca/models/ClientStats.java
```diff
@@ -0,0 +1,12 @@
...
+
+import java.util.List;
+
+/** Statistic Model for each client. Maily handle KPI and metric calculations */
```
RockyMandayam2, line 5: `Statistic Model` -> `Aggregated information` or `Aggregated statistics`?

lib-rca/src/main/java/com/facebook/openwifi/librca/models/ClientStats.java
```diff
@@ -0,0 +1,12 @@
...
+public class ClientStats {
+    /** Unique ID for each client. */
```
RockyMandayam2, line 7: Can the javadoc just say `client MAC`?

lib-rca/src/main/java/com/facebook/openwifi/librca/models/ClientStats.java
RockyMandayam2, line 1: Add copyright

lib-rca/src/main/java/com/facebook/openwifi/librca/models/InputStats.java
```diff
@@ -0,0 +1,56 @@
...
+
+/** Input Data Model */
+public class InputStats {
+    public static class RadioConfig {
+        public int channel;
+        public int channelWidth;
+        public int txPower;
+        public String phy;
+    }
+    public class Association{
```
RockyMandayam2, line 19: `public static`

lib-rca/src/main/java/com/facebook/openwifi/librca/models/InputStats.java
```diff
@@ -0,0 +1,56 @@
...
+        /** Rate information for receive/transmit data rate*/
+        public class Rate {
```
RockyMandayam2, line 21: `public static`

lib-rca/src/main/java/com/facebook/openwifi/librca/models/InputStats.java
```diff
@@ -0,0 +1,56 @@
...
+            public long bitRate;
+            public int chWidth;
+            public int mcs;
```
RockyMandayam2, line 24: Add javadoc saying it's Modulation and Coding Scheme

lib-rca/src/main/java/com/facebook/openwifi/librca/models/InputStats.java
```diff
@@ -0,0 +1,56 @@
...
+        public long activeMsRadio;
+        public long busyMsRadio;
+        public long noiseRadio;
+        public long receiveMsRadio;
+        public long transmitMsRadio;
+
+        /** Unix time in milliseconds. */
+        public long timestamp;
+    }
+    public String bssid;
```
RockyMandayam2, line 52: Add javadoc saying this is the bssid of the AP radio

lib-rca/src/main/java/com/facebook/openwifi/librca/models/InputStats.java
```diff
@@ -0,0 +1,56 @@
...
+    public String station;
```
RockyMandayam2, line 53: Add javadoc saying this is the client MAC

lib-rca/src/main/java/com/facebook/openwifi/librca/models/LinkStats.java
```diff
@@ -0,0 +1,60 @@
...
+ * Aggregation Statistics Model of InputStats. Aggregate by bssid, station and RadioConfig.
+ */
+public class LinkStats {
+    public static class RadioConfig {
+        public int channel;
+        public int channelWidth;
+        public int txPower;
+        public String phy;
+    }
+    public class Association{
```
RockyMandayam2, line 23: `public static`

lib-rca/src/main/java/com/facebook/openwifi/librca/models/LinkStats.java
```diff
@@ -0,0 +1,60 @@
...
+        /** Rate information for receive/transmit data rate*/
+        public class Rate {
```
RockyMandayam2, line 25: `public static`

lib-rca/src/main/java/com/facebook/openwifi/librca/models/LinkStats.java
RockyMandayam2, line 23: If this `Association` is the same as the `Association` in `InputStats` I think we should extract the `Association` class out separately and avoid duplicate code.

lib-rca/src/main/java/com/facebook/openwifi/librca/models/LinkStats.java
```diff
@@ -0,0 +1,60 @@
...
+
+package com.facebook.openwifi.librca.models;
+
+import java.util.List;
+
+/** 
+ * Aggregation Statistics Model of InputStats. Aggregate by bssid, station and RadioConfig.
+ */
+public class LinkStats {
+    public static class RadioConfig {
```
RockyMandayam2, line 17: I think we should extract the `RadioConfig` class out separately and avoid duplicate code (it's the same as `InputStats.RadioConfig`

lib-rca/src/main/java/com/facebook/openwifi/librca/models/ClientStats.java
```diff
@@ -0,0 +1,12 @@
...
+    String station;
+    
+    /** LinkStats that are of the same station(client). Because of disconnection issues, a client might be connected to several BSSIDs. */
```
RockyMandayam2, line 10: I think this javadoc is a bit misleading - it seems to imply that disconnection issues are why there may be multiple `LinkStats` objects. I think you could just leave it at `LinkStats that are of the same station(client)` and if someone wanted more info they could look at the javadoc for `LinkStats`.

The `LinkStats` are aggregated by bssid, station, and radio config. This means each `LinkStats` will have a unique bssid, station, and radio config. Therefore, if the radio config changes (e.g., different tx power), there will be multiple `LinkStats` created.

RockyMandayam2 (changes requested):

lib-rca/src/main/java/com/facebook/openwifi/librca/models/ClientStats.java
halmehmood: I agree that we should remove "Because of disconnection issues,"

lib-rca/src/main/java/com/facebook/openwifi/librca/models/ClientStats.java
halmehmood: I agree, client MAC is a better description.

lib-rca/src/main/java/com/facebook/openwifi/librca/models/InputStats.java
```diff
@@ -0,0 +1,56 @@
...
+        public long txPackets;
+        public Rate txRate;
+        public long txRetries;
+        public int ackSignal;
+        public int ackSignalAvg;
+        public long txPacketsCounters;
+        public long txErrorsCounters;
+        public long txDroppedCounters;
+        public long activeMsRadio;
+        public long busyMsRadio;
```
halmehmood, line 44: Maybe add a comment here as well mentioning that the metrics below are from the radio the client was associated to

lib-rca/src/main/java/com/facebook/openwifi/librca/models/InputStats.java
```diff
@@ -0,0 +1,56 @@
...
+        public Rate rxRate;
+        public long txBytes;
+        public long txDuration;
+        public long txFailed;
+        public long txPackets;
+        public Rate txRate;
+        public long txRetries;
+        public int ackSignal;
+        public int ackSignalAvg;
+        public long txPacketsCounters;
```
halmehmood, line 40: Maybe add a comment mentioning that these counters are for the wireless interface as a whole

halmehmood (commented):
The format I had suggested where the radio metrics and counters metrics go in the same struct as the station metrics (like you have defined in Association) was for LinkStats. We should not expect the input stats to be in that format. This format is for our RCA usage. For getting data from an external source, we should have a more industry-standard-ish format where an AP/interface has several radios, each radio has several BSSIDs and each BSSID has several associations and the corresponding stats go under each category.

lib-rca/src/main/java/com/facebook/openwifi/librca/models/ClientStats.java
zhiqiand: No, inputstats is not aggregated.

RockyMandayam2: Maybe we can separate that from the other classes by putting it in a different package called `inputs`? Or maybe they can go here under different sub-packages?

lib-rca/src/main/java/com/facebook/openwifi/librca/models/LinkStats.java
zhiqiand: Which one do you suggest? A separate file or just put it in the same file?

RockyMandayam2: We discussed offline

lib-rca/src/main/java/com/facebook/openwifi/librca/stats/LinkStats.java
```diff
@@ -0,0 +1,73 @@
...
+		public int ackSignal;
+		public int ackSignalAvg;
+
+		// The metrics below are from Interface the client was connected to
+		public long txPacketsCounters;
+		public long txErrorsCounters;
+		public long txDroppedCounters;
+
+		// The metrics below are from the radio the client was associated to.
+		// These counters are for the wireless interface as a whole.
```
halmehmood, line 53: I think this line should be below line 47

lib-rca/src/main/java/com/facebook/openwifi/librca/stats/inputs/InputStats.java
```diff
@@ -0,0 +1,74 @@
...
+			public long tx_failed;
+			public long tx_offset;
+			public long tx_packets;
+			public Rate tx_rate;
+			public long tx_retries;
+			public int ack_signal;
+			public int ack_signal_avg;
+		}
+		public Association[] associations;
+		public Radio radio;
```
halmehmood, line 54: The Radio class does not need to be under SSID. Instead SSID can be under Radio.

halmehmood (commented):
Thanks for addressing my comments. Left a couple of minor comments here.

lib-rca/src/main/java/com/facebook/openwifi/librca/stats/inputs/InputStats.java
```diff
@@ -0,0 +1,74 @@
...
+ */
+
+package com.facebook.openwifi.librca.stats.inputs;
+
+/** Input Data Model */
+public class InputStats {
+    /** Radio parameters */
+    public static class Radio {
+        public long active_ms;
+        public long busy_ms;
```
RockyMandayam2, line 16: Can you change all the snake case in all these files to camel case? It does not have to be the same as field names in OpenWiFi

lib-rca/src/main/java/com/facebook/openwifi/librca/stats/inputs/InputStats.java
```diff
@@ -0,0 +1,74 @@
...
+		public static class Association {
+			/** Modulation and Coding Scheme */
+			public static class Rate {
+				public long bitrate;
+				public int chwidth;
+				public int mcs;
+			}
+			// bssid of the AP radio
+			public String bssid;
+			// client MAC
```
RockyMandayam2, line 35: Nit: Change this to a one-line javadoc: `/** client MAC */`

lib-rca/src/main/java/com/facebook/openwifi/librca/stats/inputs/InputStats.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+	public static class SSID {
+		public static class Association {
+			/** Modulation and Coding Scheme */
+			public static class Rate {
+				public long bitrate;
+				public int chwidth;
+				public int mcs;
+			}
+			// bssid of the AP radio
```
RockyMandayam2, line 33: Nit: Change this to a one-line javadoc: `/** bssid of the AP radio */`

lib-rca/src/main/java/com/facebook/openwifi/librca/stats/inputs/InputStats.java
```diff
@@ -0,0 +1,74 @@
...
+			public long tx_retries;
+			public int ack_signal;
+			public int ack_signal_avg;
+		}
+		public Association[] associations;
+		public Radio radio;
+	}
+
+    /** Counters are for the wireless interface as a whole. */
+	public static class Counters {
```
RockyMandayam2, line 58: I just noticed this - in the openwifi State, a counter is associated with an interface. However, we don't have an `Interface` in `InputStats`. How should we resolve this? @halmehmood

lib-rca/src/main/java/com/facebook/openwifi/librca/stats/inputs/InputStats.java
```diff
@@ -0,0 +1,74 @@
...
+		public long tx_bytes;
+		public long tx_packets;
+		public long tx_errors;
+		public long tx_dropped;
+	}
+
+	public SSID[] ssids;
+	public Counters counters;
+
+	// Unix time in milliseconds.
```
RockyMandayam2, line 72: Nit: Change this to a one-line javadoc: `/** Unit time in milliseconds */`

RockyMandayam2 (changes requested):

lib-rca/src/main/java/com/facebook/openwifi/librca/stats/inputs/InputStats.java
zhiqiand: I can add `Interface` here but the structure will be a bit nested. @halmehmood, any thoughts?

PR #127, opened 2022-11-12
Fix owprov API handler /inventory/<serial>?rrmSettings=true

elludraon:

Starting to see `{"rrm":"no"}` returned from this endpoint, don't know why. Don't treat this as a deserialization error.

Ref: https://github.com/Telecominfraproject/wlan-cloud-owprov/blob/6aee513a4511b07f651708786e5fdf02862939ca/src/RESTAPI/RESTAPI_inventory_handler.cpp#L72-L73

### DIFF
--- a/lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/UCentralClient.java
+++ b/lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/UCentralClient.java
@@ -690,6 +690,19 @@ public class UCentralClient {
 		try {
 			return gson.fromJson(response.getBody(), RRMDetails.class);
 		} catch (JsonSyntaxException e) {
+			// catch strings like "no", "inherit", "invalid" (???)
+			JSONObject respBody;
+			try {
+				respBody = new JSONObject(response.getBody());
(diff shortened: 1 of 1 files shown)


pohanhf (approved):

lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/UCentralClient.java
```diff
@@ -690,6 +690,19 @@ public RRMDetails getProvInventoryRRMDetails(String serialNumber) {
...
+			JSONObject respBody;
+			try {
+				respBody = new JSONObject(response.getBody());
+				respBody.getString("rrm");
+				logger.error(
+					"RRMDetails returned unexpected string body: {}",
+					respBody
+				);
+				return null;
+			} catch (JSONException e2) { /* ignore and fall through */}
```
wuweicaifb, line 703: Add some error logs in this catch block?

elludraon: This block was only checking if the body was a different type which isn't expected anyway, which is why we ignore any error here.

PR #128, opened 2022-11-12
Client steering notes, docs, fixes

elludraon:
Fill in a bit more of the initial client steering implementation, but still not enough to be functional.
- Add detailed TODO comments on commands which need to be sent to APs/clients
- Update docs at `ALGORITHMS.md`
- Parse `RMEnabledCapabilities` IE needed to determine 802.11k support
- Fix `backoffTimeNs` argument and change to `backoffTimeSec` (switch to more reasonable time unit)

### DIFF
--- a/owrrm/src/main/java/com/facebook/openwifi/rrm/optimizers/clientsteering/ClientSteeringOptimizer.java
+++ b/owrrm/src/main/java/com/facebook/openwifi/rrm/optimizers/clientsteering/ClientSteeringOptimizer.java
@@ -18,7 +18,7 @@ import com.facebook.openwifi.rrm.modules.Modeler.DataModel;
 public abstract class ClientSteeringOptimizer {
 	// TODO call upon triggers, not only via one-off or period runs
 
-	/** Represents client steering actions an AP Can take */
+	/** Represents client steering actions an AP can take */
 	public static enum CLIENT_STEERING_ACTIONS {
 		/** Steer from 2G to 5G/6G */
 		STEER_UP,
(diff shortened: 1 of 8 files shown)


owrrm/src/main/java/com/facebook/openwifi/rrm/optimizers/clientsteering/ClientSteeringOptimizer.java
```diff
@@ -28,6 +28,47 @@ public static enum CLIENT_STEERING_ACTIONS {
...
+		public static final int DISASSOC_LOW_ACK = 34;
+		public static final int MESH_PEERING_CANCELLED = 52;
+		public static final int MESH_MAX_PEERS = 53;
+		public static final int MESH_CONFIG_POLICY_VIOLATION = 54;
+		public static final int MESH_CLOSE_RCVD = 55;
+		public static final int MESH_MAX_RETRIES = 56;
+		public static final int MESH_CONFIRM_TIMEOUT = 57;
+		public static final int MESH_INVALID_GTK = 58;
+		public static final int MESH_INCONSISTENT_PARAMS = 59;
+		public static final int MESH_INVALID_SECURITY_CAP = 60;
```
wuweicaifb, line 69: How about error code after 60?

wuweicaifb (approved):

owrrm/src/main/java/com/facebook/openwifi/rrm/optimizers/clientsteering/ClientSteeringOptimizer.java
elludraon: Thanks, missed those.

PR #129, opened 2022-11-21
Remove device config schema copy in ConfigManager

elludraon:

Remove persistent member `ConfigManager.DeviceData.config` as it is fetched every time anyway, which reduces overall memory requirements for the service. (This is replaced with local/temporary variables.)


pohanhf (approved):
LGTM.

PR #130, opened 2022-11-22
Add UCentralSchema and convert Modeler.DataModel.latestDeviceStatusRadios

elludraon:
Convert `Modeler.DataModel.latestDeviceStatusRadios` value type from `JsonArray` to `UCentralSchema.Radio`, which brings some memory savings.

Supersedes #110, but omits the full switch of the `UCentralApConfiguration` underlying structure (still uses `JsonObject` instead of the new `UCentralSchema`) as that would require larger changes.

### DIFF
--- a/lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/models/ap/UCentralSchema.java
+++ b/lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/models/ap/UCentralSchema.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ * All rights reserved.
+ *
+ * This source code is licensed under the BSD-style license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+package com.facebook.openwifi.cloudsdk.models.ap;
+
+import java.util.List;
+
+import com.google.gson.JsonPrimitive;
+import com.google.gson.annotations.SerializedName;
+
+public class UCentralSchema {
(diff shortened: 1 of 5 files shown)


lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/models/ap/UCentralSchema.java
```diff
@@ -0,0 +1,91 @@
...
+ */
+
+package com.facebook.openwifi.cloudsdk.models.ap;
+
+import java.util.List;
+
+import com.google.gson.JsonPrimitive;
+import com.google.gson.annotations.SerializedName;
+
+public class UCentralSchema {
```
pohanhf, line 16: Not sure whether it will be useful or not: do you think it might be helpful to put the schema link here so people can know what we are referring to quickly (so they don't need to be like us searching around all the time)? Thanks.

pohanhf (approved):
Thanks for the PR! I like this change not only because it saves memory compared to JsonArray but also it makes the structure (inside the previous Json stuff) easy to follow. Also, instead of checking the schema all the time from the website, people can easily get the same information from the UCentralSchema.java file. Thanks :).

lib-cloudsdk/src/main/java/com/facebook/openwifi/cloudsdk/models/ap/UCentralSchema.java
elludraon: Good idea, will follow up separately.

